PR #7, opened 2015-02-04
Update apc.ini.j2

relaxnow:


thijskh:
Fixed in 28de93defd63623dd556914ad6eb36cd6e908a07

relaxnow:
Actually that commit did not increase the shared memory size.

PR #25, opened 2015-11-11
DIY & misc fixes

baszoetekouw:
A number of fixes in the DIY script.  Main change is introduction of the variable "nginx_backend_ip", which describes the ip nginx connects to in the backend.  Typically, this will be 127.0.0.1, but for DIY, it is set to 127.0.0.2 as we are not using a oad balancer (so no HAproxy, and nginx and apache on the same machine).


baszoetekouw:
Hmm, please don;t merge this yet; something is till wrong with grouper/teams stems

baszoetekouw:
Should be fine now.

roles/php/tasks/main.yml
```diff
@@ -15,6 +15,7 @@
      - php-soap
      - php-xml
      - php-mcrypt
+     - php-gd
```
thijskh: Why is php-gd needed?

roles/teams/tasks/main.yml
```diff
@@ -57,7 +56,6 @@
   changed_when: "gsh_output.stderr != ''"
   args:
     chdir: /opt/www/grouper-shell
-  when: env == 'vm'
```
thijskh: We're now running grouper provisioning on every teams deploy. Is that necessary? I think the "when" was added precisely to avoid messing with grouper provisioning in already installed envs.

roles/php/tasks/main.yml
baszoetekouw: I'm not entirely sure, but EB refused to install without it.

roles/teams/tasks/main.yml
baszoetekouw: The script itself checks whether the stem already exists, and if so, does nothing.  Also, creating additional stems, if it would happend by accident, doesn't matter for Grouper or Temas operations.  A missing stem, otoh, does matter quite a lot.

thijskh: Alright, I'll merge it then, if it gives rise to problems we should fix those there.

PR #26, opened 2015-11-20
Updated secrets to allow for entering secrets file using command line

surfnet-niels:
The updated script allows the location of the secrets file to be entered on the command line.
If no location is provided, the default location is used


thijskh:
Thanks, I've merged it but not before squashing the last commit (because it deletes a file that was added in the same PR, seemed like needless churn in the repo).

PR #32, opened 2015-12-02
Set develop flag to false by default

arothuis:
The develop flag is used for development-specific tasks/roles and is set to true via --extra-vars. This change allows Ansible to default to `develop = false` if it is not given as an extra-var.

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -1,4 +1,6 @@
 ---
+develop: false
+
 timezone: Europe/Amsterdam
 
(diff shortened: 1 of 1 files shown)


DRvanR:
Please add a bit of documentation as to what it is and why it should be false by default. Once that has been done, :+1:

arothuis:
@DRvanR, @rjkip

DRvanR:
:+1:

rjkip:
:+1:

PR #37, opened 2015-12-23
Adds metadata feature toggle

arothuis:
A feature toggle is added to EngineBlock so that we can enable/disable the metadata API. This needs to be added to the provisioned configuration.


arothuis:
@rjkip

rjkip:
:+1:

PR #40, opened 2016-02-03
Updates EB symfony log and cache dirs

arothuis:
In production, syslog and the default Symfony directories are used and do not need to be created.

For development specific log and cache directories are used: `/tmp/engineblock/cache` and `/tmp/engineblock/log`.

See: https://github.com/OpenConext/OpenConext-engineblock/pull/264


DRvanR:
:+1:

PR #42, opened 2016-02-04
Correct quoting of Attribute Aggregation password

DRvanR:
Added missing quote to ensure Engineblock can be booted.


DRvanR:
@thijskh @oharsta @quartje

thijskh:
:+1:

quartje:
:+1:

PR #52, opened 2016-03-14
Install composer for engineblock 5

DRvanR:
EngineBlock requires composer in order to be able to execute scripts that allow the local environment to influence the application (e.g. configuration through local ini files). These changes install composer through yum.


DRvanR:
@quartje

PR #53, opened 2016-03-15
Enable EngineBlock API vhost, expose SYMFONY_ENV, dev env fixes

DRvanR:
- Enabled updated engine API vhost
- Cleaned up EB5 provisioning
- Ensure EB5 scripts that run run using the correct SYMFONY_ENV


DRvanR:
@quartje for some reason having this setup makes basic auth fail, meaning the PUSH api of engineblock does not work (the API itself does, as it reports through the heartbeat). When debugging on the engineblock side of things it seems that no username and password (basic auth) is received... any ideas?

DRvanR:
fixed it, seems some solutions only work with more recent versions of apache2.4. Oh well, this seems to do the trick

DRvanR:
@quartje please review and merge if applicable.

PR #60, opened 2016-04-07
Allow external database connections in dev env

DRvanR:
- add iptables rule for port 3306
- set bind-address to 0.0.0.0 in dev env


DRvanR:
@quartje

quartje:
:+1:

PR #62, opened 2016-04-07
Added libselinux-python to standerd deploy to make sure python can interact with SElinux

surfnet-niels:


baszoetekouw:
Why do we need libselinux-python in the VM?  Surely, to do ansible SE-linux stuff, we only need the python SELinux libs locally?

surfnet-niels:
locally as in on the deploy host?

SE linux needs to be set in permissive mode and some other task are needed, see https://github.com/OpenConext/OpenConext-deploy/blob/master/roles/common/tasks/main.yml#L23

These will not work without libselinux-python

baszoetekouw:
As far as I know, ansible only needs local dependencies (on the deploy host).  It wraps everything up in a self-contained python script that is copied to the remote host and executed there.  You don't need any dependencies for ansible on the remote hosts (except for pyhton itself).

pmeulen:
Note that some Ansible modules do need some (python) dependencies to be present to work on the remote host. This is because the "self-contained python script" needs them while it is being executed on the remote host. The selinux module indeed requires libselinux-python: http://docs.ansible.com/ansible/selinux_module.html#requirements-on-host-that-executes-module

To see what is being executed on the remote you can set ANSIBLE_KEEP_REMOTE_FILES=1 in the environment on you local host, this retains the python module that was executed by Ansible on the remote in ~/.ansible/tmp/ on the remote.

PR #71, opened 2016-04-13
Add vhost_name to haproxy applications list.

thijskh:
haproxy.cfg decides based on the vhost name which backend to use, but not
all vhost names match the application name; therefore this needs to be
specified explictly. This fixes engine-api, authz, authz-admin, and
authz-playground.


quartje:
:+1: 
Merged by hand

PR #73, opened 2016-05-10
Fix ansible error reported by some OpenConext users

baszoetekouw:
No idea why this fails (works for me), but it has been reported by two separated OpenConext users...


thijskh:
Doesn't seem to hurt to change it then...

PR #80, opened 2016-07-21
Dev env provisionig

DRvanR:
- [X] give sensible names to vms for Virtualbox UI
- [X] use `env=vm` for ansible provisioning
- [X] ensure database migrations run in the correct order
    this is needed because the current dump no longer includes all tables (notably, `sso_provider_roles` is missing)
- [ ] ensure haproxy health checks are disabled in dev env (see [here][1])

[1]: https://github.com/OpenConext/OpenConext-deploy/commit/b27a8441979b118a037475ae32af2679930dd022#diff-20f5428347b213f92e8081ea12666683L41


quartje:
:+1:

PR #85, opened 2016-08-10
EBHost setup using only EngineBlock and Janus. 

muisit:
This setup disables Janus authentication through EB and mujina due to mujina not being deployed.

The EBHost environment supposedly is a base for creating a 2-stack EBHost DB-linked environment and a 2-stack EBHost DB-linked environment with load-balancing later on.


quartje:
Thanks for this work! I will do some testing before commenting furher and see how it works out in our test environment.

PR #86, opened 2016-08-11
Specify which virtual boxes are configured

arothuis:
This fixes an issue where lb_centos7 was getting the configuration for apps_centos7, causing conflicting names for the VMs in VirtualBox.


arothuis:
The following output illustrates the issue, `lb_centos7`'s name is set to `OpenConext Engineblock Apps`:
```
Bringing machine 'lb_centos7' up with 'virtualbox' provider...
Bringing machine 'apps_centos7' up with 'virtualbox' provider...
==> lb_centos7: Importing base box 'CentOS-7.0'...
==> lb_centos7: Matching MAC address for NAT networking...
==> lb_centos7: Setting the name of the VM: OpenConext Engineblock Apps
```

I don't think the failing build is related to this change as this only changes the settings for VirtualBox VMs. Provisioning, during which Travis' error occurs, should remain untouched.

PR #89, opened 2016-08-25
Add parameter and configuration for error response time padding

arothuis:
#### Related PR
[EngineBlock PR #328](https://github.com/OpenConext/OpenConext-engineblock/pull/328)

#### Implements issue
[127989925](https://www.pivotaltracker.com/story/show/127989925)

### DIFF
--- a/environments/vm/secrets/vm.yml
+++ b/environments/vm/secrets/vm.yml
@@ -26,6 +26,9 @@ engine_janus_secret: secret
 
 engine_ldap_password: secret
 
+# Minimum execution time in milliseconds when response is invalid
+engine_minimum_execution_time_on_invalid_response: 5000
+
 # below is the result of  slappasswd -s secret  -h {SSHA}
 # todo: replace the manual use of slapdpassword by some automated ansible
 # magic that runs:
@@ -64,7 +67,6 @@ authz_admin_client_secret: secret
 authzplayground_springapp_authz_client_secret: secret
 authzplayground_springapp_oidc_client_secret: secret
 
-engine_padded_error_response_time: 5
 
 pdp_password: secret
 pdp_client_secret: secret
--- a/environments/vm/secrets/vm.yml
+++ b/environments/vm/secrets/vm.yml
@@ -26,9 +26,6 @@ engine_janus_secret: secret
 
 engine_ldap_password: secret
 
-# Minimum execution time in milliseconds when response is invalid
-engine_minimum_execution_time_on_invalid_response: 5000
-
 # below is the result of  slappasswd -s secret  -h {SSHA}
 # todo: replace the manual use of slapdpassword by some automated ansible
 # magic that runs:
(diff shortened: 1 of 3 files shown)


arothuis:
@DRvanR

Not sure about where to place the template parameter. For now, I figured as it is related to security, I've put it in the secrets file.

environments/vm/secrets/vm.yml
```diff
@@ -26,6 +26,9 @@ engine_janus_secret: secret
 
 engine_ldap_password: secret
 
+# Minimum execution time in milliseconds when response is invalid
+engine_minimum_execution_time_on_invalid_response: 5000
```
DRvanR: Is this actually required?

environments/vm/secrets/vm.yml
```diff
@@ -64,6 +67,7 @@ authz_admin_client_secret: secret
 authzplayground_springapp_authz_client_secret: secret
 authzplayground_springapp_oidc_client_secret: secret
 
+
```
DRvanR: can go?

roles/engineblock/defaults/main.yml
```diff
@@ -51,6 +51,9 @@ engine_api_profile_password: secret
 
 engine_engine_idp_baseurlpath: "https://{{ engine_domain }}/simplesaml/"
 
+# Minimum execution time in milliseconds when response is invalid
+engine_minimum_execution_time_on_invalid_response: 5000
```
DRvanR: the more I see it... `...on_invalid_received_response`? I mean it's not that EB gives an invalid response, it's that we get one that is deemed invalid...

arothuis: Hmm. `...when_invalid_response_received`?

Note: if this changes the related PR in EB should also be changed.

environments/vm/secrets/vm.yml
arothuis: What do you mean by 'this'? Regarding placement in the secrets file, see my general comment above. We could also hard-code it in `engineblock.ini.j2`, but that would make it less transparently configurable IMHO.

roles/engineblock/defaults/main.yml
DRvanR: yes I know :wink:

Would prefer `...in_invalid_received_reponse` - the received response is deemed invalid by EB, it does not have to be invalid per se...

environments/vm/secrets/vm.yml
DRvanR: literally, this change :wink:

Why is it needed? The default is in `roles/engineblock/defaults/main.yml`, this is a VM specific override that is the exact same, what is the point of that?

arothuis: Ah, I see. I wanted it to be configured more explicitly, I guess. But you're right: we don't need it for VM.

It would be for production to override this when necessary.

PR #90, opened 2016-09-05
Disable sendfile for development

arothuis:
This mitigates the issue on certain VMs where files sent by the webserver are corrupted because of illegal characters and/or truncation.

_This is a known, long-running issue when using [VirtualBox](https://www.virtualbox.org/ticket/9069)._


arothuis:
@thijskh, @quartje

thijskh:
So this is related to the filesystem used specifically in the Virtualbox-setup. Fine with merging.

PR #91, opened 2016-09-15
Let Grouper uses the correct database name when provisioning

DRvanR:
The additional `i` led to a failed provisioning on the dev env.


DRvanR:
@quartje

PR #98, opened 2016-09-20
Use /dev/urandom as random source instead of /dev/random

baszoetekouw:
This patch changes the random source of the java programs to /dev/urandom (instead of the default /dev/random). This fixes issues with the startup time of the java apps (see OpenConext/OpenConext-voot#3), while sacrificing no security. /dev/urandom is s perfectly fine source for random data, and as even openssl uses /dev/urandom for seeding its PRNG, it should also be fine for our java apps.

See also http://www.2uo.de/myths-about-urandom/, http://sockpuppet.org/blog/2014/02/25/safely-generate-random-numbers/

### DIFF
--- a/roles/aa/vars/main.yml
+++ b/roles/aa/vars/main.yml
@@ -12,6 +12,7 @@ springapp_local_jar: "{{ aa_local_jar }}"
 springapp_debug: "{{ aa_debug }}"
 springapp_debug_port: 1198
 springapp_heapsize: "512m"
+springapp_random_source: "file:///dev/urandom"
 
 js_error_reporting_access_token: "{{ attribute_aggregator_js_access_token }}"
 js_error_enabled: "true"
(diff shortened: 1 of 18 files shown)


thijskh:
@pmeulen @oharsta

baszoetekouw:
It seems Grouper suffers from the same problem.  Without sufficient entropy, the task `grouper : initialize` hangs.  Adding ```-Djava.security.egd=file:///dev/urandom``` to the java call in `/opt/www/grouper-shell/bin/gsh` fixes that.  
I'll see if I can add that to this patch.

baszoetekouw:
I'm not sure what the best plac eit to put the config for the rnadom source for java apps: should I put it in each of the roles' defaults, in the environment's `group_vars`, or in the global `group_vars/java-apps.yml`?

Obviously, the hard-coded references to /dev/urandom in the Teams and Mujina tasks file still need to be replaced by a reference to a variable that defaults to `/dev/urandom`.

baszoetekouw:
@oharsta Do you know if there is a way to set this as the global default for java on a machine-level?  That would be quite a bit easier than setting it manually fo each java-call.

oharsta:
I don't know a way to set this globally. Best thing to handle this through ansible in the group_vars/java-apps.yml

mrvanes:
https://wiki.apache.org/tomcat/HowTo/FasterStartUp#Entropy_Source

quartje:
Handling it in the group_vars/java-apps.yml means this group always needs to be present. Leaving it in the individual roles defaults is therefor safer.

PR #101, opened 2016-09-26
Small fixes

baszoetekouw:


roles/haproxy/tasks/single-instance.yml
```diff
@@ -4,24 +4,31 @@
...
   notify:
     - restart haproxy
 
 - name: Create haproxy SSL key directory
-  file: dest=/etc/pki/haproxy/ state=directory owner=haproxy group=haproxy mode=0700
+  file: "dest=/etc/pki/haproxy/ state=directory owner=haproxy group=haproxy mode=0700"
 
 - name: Create combined key and certificate file for HAproxy 
-  shell: cat {{ tls.cert_private_path }}/{{ item.key_name }}  {{ tls.cert_path }}/{{ item.crt_name }} > /etc/pki/haproxy/{{ item.name }}_haproxy.pem
+  shell: "umask 600; cat {{ tls.cert_private_path }}/{{ item.key_name }}  {{ tls.cert_path }}/{{ item.crt_name }} > /tmp/{{ item.name }}_haproxy.pem"
```
thijskh: Temp file race...?

baszoetekouw: Yeah, the temp file thing is quite hackish.  As @pmeulen pointed out, it would be nicer to put the ssl cert into a variabel rather than a file, so that it more easily be tracked by ansible.  Writing it to a file would than simply be a matter of using ```copy```'s ```content``` option.

I'll see if I can fix this up soonish.

PR #114, opened 2016-10-10
Global instance name

thijskh:
Allows to globally set the name of this installation, e.g. "OzonConext" or "SURFconext". Not all apps will yet pick this up because they lack a global configuration parameter for the name.


thijskh:
@baszoetekouw @surfnet-niels

quartje:
I have tested it by installing a complete new instance from the template. Works fine!

PR #116, opened 2016-12-02
Feature/encrypt haproxy backend traffic

quartje:
After this change the Apache backends can be served over https, so the traffic between Haproxy and Apache is encrypted. 
An approach has been chosen where one certifcate/key pair is used for all apache backends. This can be either a certificate obtained by a public CA, or a certificate signed by your own CA. 
Enabling this encryption is switchable. It's switched off by default. 
The Docker environment used for Travis testing has https backends enabled for automatic testing.


thijskh:
:+1:

If you want to make it even more fancy you make it two-way authenticated with an SSL client cert aswell. But... baby steps.

quartje:
That would make it fancy indeed. But for now this seems enough to protect the traffic between loadbalancer and webserver backends.

PR #119, opened 2016-12-21
Bring provision-vm.yml and provision-docker.yml in sync

baszoetekouw:
Provision-vm,yml and provision-docker.yml are almost identical; this change makes this obvious.

As a consequence, we are now deploying oidc in docker/travis.  This required a minor change to the vm_only_oidc task to make it reentrant and safe to run twice.

### DIFF
--- a/provision-docker.yml
+++ b/provision-docker.yml
@@ -11,9 +11,9 @@
   gather_facts: yes
   become: true
   roles:
-    - common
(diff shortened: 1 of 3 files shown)


baszoetekouw:
Hmm, the oidc thing doesn't seem to work like this, unfortunately... looking for a nice way to solve this.

PR #120, opened 2017-01-17
Feature/minimal install

muisit:
Moved the single host installation configuration to the main provision-template.yml playbook using the 'extra-var' minimal_install.
Usage:
ansible-playbook -i ..... --extra-vars="secret..... minimal_install=1"

This installs only EngineBlock and the Service-Registry and skips the HAProxy load balancer.


quartje:
Seems to be working fine! I'll add some small improvements after merging.

PR #121, opened 2017-01-23
Modify EngineBlock PDP client base URL for new PDP client

arothuis:
The modifies the base URI to be used by the new Guzzle 6-based PDP client.

See: OpenConext/OpenConext-engineblock#377


DRvanR:
More specifically, see this: https://github.com/OpenConext/OpenConext-engineblock/pull/377/files#diff-ed08d04023a5acaadc6a599cc808020fR44

:+1:

PR #123, opened 2017-01-30
PDP host and path configuation for EB 5.2

DRvanR:
Allows the configuration of the PDP host and path for engineblock 5.2 in a backwards compatible way


DRvanR:
@thijskh @quartje thoughts?

PR #124, opened 2017-02-01
Feature/minimal install

quartje:
With this PR it becomes possible to install a minimal OpenConext instance, with only engineblock, profile and serviceregistry. 
No IdP is present. Login to serviceregistry is only possible with a local admin account.


thijskh:
LGTM :+1:

PR #125, opened 2017-02-08
Use same playbook for travis build and VM installs

baszoetekouw:
To make sure the VM installs correctly, it is preferable for Travis to us the same playbook ans the VM install.  The playbooks themselves were already almost identical, apart from the OIDC role, which was missing from the Travis build.  A few minor tweaks were necessary to get things to work in Docker, but nothing major.


thijskh:
Thanks, looks good to me!

PR #126, opened 2017-02-08
Move Travis build instructions into a proper shell script

baszoetekouw:
The current travis build commands (in ``.travis.yml``) are quite hard to read, edit and debug.  This PR moves all of the build stuff into a proper shell script, which is called from ``.travis.yml``.  This simplifies the build process quite a lot, because we are no longer dependent on unreadable shell oneliners wrapped in  yaml.


thijskh:
:+1:

PR #127, opened 2017-03-03
sendmail to postfix

thijskh:


quartje:
:thumbsup: Seems to be working fine, Sendmail is replaced without problems by Postfix.

PR #129, opened 2017-04-25
Remove both ldap and Grouper.

quartje:
Both are not needed anymore, since al functionality is provided by Teams now


thijskh:
Should probably also remove tomcat from default java-common, and move to oidc, so you only get tomcat if you install oidc?

Tomcat6 stuff can be removed entirely I think.

roles/diyidp/templates/frontpage.definition.json.j2
```diff
@@ -26,9 +26,6 @@
 	"required": {
 		"en": "Required"
 	},
-	"required_ldap": {
-		"en": "Required for LDAP"
-	},
 	"required_radius": {
```
thijskh: Eh, this is a simplesaml template.

roles/janus/tasks/simplesamlphp.yml
```diff
@@ -56,7 +53,6 @@
     - "{{ janus_ssp_modules_dir }}/exampleattributeserver"
     - "{{ janus_ssp_modules_dir }}/expirycheck"
     - "{{ janus_ssp_modules_dir }}/InfoCard"
-    - "{{ janus_ssp_modules_dir }}/ldap"
     - "{{ janus_ssp_modules_dir }}/logpeek"
     - "{{ janus_ssp_modules_dir }}/memcacheMonitor"
     - "{{ janus_ssp_modules_dir }}/metaedit"
```
thijskh: So we _stop_ _removing_ these modules? Makes no sense to me.

roles/vm_only_provision_eb_sr/templates/serviceregistry.sql.j2
```diff
@@ -577,7 +575,6 @@ INSERT INTO `janus__metadata` (`connectionRevisionId`, `key`, `value`, `created`
 INSERT INTO `janus__metadata` (`connectionRevisionId`, `key`, `value`, `created`, `ip`) VALUES (7,'redirect.sign','','2012-06-13T08:02:29+02:00','172.18.5.1');
 {% if ( not minimal_install ) %}
 INSERT INTO `janus__metadata` (`connectionRevisionId`, `key`, `value`, `created`, `ip`) VALUES (8,'AssertionConsumerService:0:Binding','urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST','2012-06-19T09:44:38+02:00','172.18.5.1');
-INSERT INTO `janus__metadata` (`connectionRevisionId`, `key`, `value`, `created`, `ip`) VALUES (8,'AssertionConsumerService:0:Location','https://grouper.{{ base_domain }}/grouper/Shibboleth.sso/SAML2/POST','2012-06-19T09:44:38+02:00','172.18.5.1');
 INSERT INTO `janus__metadata` (`connectionRevisionId`, `key`, `value`, `created`, `ip`) VALUES (8,'contacts:0:contactType','technical','2012-06-19T09:44:38+02:00','172.18.5.1');
 INSERT INTO `janus__metadata` (`connectionRevisionId`, `key`, `value`, `created`, `ip`) VALUES (8,'contacts:0:emailAddress','','2012-06-19T09:44:38+02:00','172.18.5.1');
```
thijskh: These changes only remove half the data related to the grouper entity, creating foreign key problems (and cruft).

quartje: You're right: I've removed the rest as well

roles/php56fpm/tasks/main.yml
```diff
@@ -21,7 +21,6 @@
   - php-cli
   - php-pecl-apcu
   - php-curl
-  - php-ldap
```
quartje: The current engineblock still needs php-ldap. We'll remove it after the EB release without ldap is released

PR #139, opened 2017-06-15
Rename option to skip engine db migrations.

thijskh:
This should make it more evident what it does.


quartje:
I had committed the same fix later, without first checking the pull requests ..

PR #142, opened 2017-07-19
Synced Vagrant dev vm group_vars with Docker

MKodde:
While rebuilding my dev environment I ran into some provisioning issues.

A quick look in the logs showed an update to `oidc`. As a quick fix I copied these settings to the vagrant vm `group_vars`.

Can someone who knows the implications of this change review this PR?


thijskh:
The vars that you add are indeed needed for the latest version of tbe attribute aggregator.

PR #148, opened 2017-10-23
Mongo: lineinfile ansible version issue

MKodde:
The commit below fixes failing provisioning of the Mongo task. 

I'm using an older version of Ansible (2.2.1.0) Should I upgrade to a newer version of ansible or is this proposed changed an approved solution?

The Ansible [docs](http://docs.ansible.com/ansible/latest/lineinfile_module.html)


quartje:
I saw the PR after I pushed the fix myself. I don't understand why Ansible has added a parameter with a different name, doing the same. Anyway, it should be fixed now.

PR #149, opened 2017-10-25
Add manage config parameters to vm.yml

MKodde:
**I'm trying to solve this problem:**
A [template](https://github.com/SURFnet/sp-dashboard/pull/43/commits/be1a7feb177745b635a474a9a773ab0c1b6a325c) in the sp-dashboard project needs to reference to Manage specific variables.

**Proposed solution**
These variables should be set in the OpenConext-deploy group vars.

I'm not a 100% sure this is the correct way to add these config parameters. Please feel free to give feedback!

### DIFF
--- a/environments/vm/group_vars/vm.yml
+++ b/environments/vm/group_vars/vm.yml
@@ -413,6 +413,7 @@ monitoring_tests:
   password: "{{ monitoring_tests_password }}"
 
 manage:
+  base_url: https://manage.{{ base_domain }}
   db_user: "{{ janus_database_user }}"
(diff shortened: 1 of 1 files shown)


environments/vm/group_vars/vm.yml
```diff
@@ -439,7 +441,7 @@ manage:
         scopes: ["READ"]
       }
     - {
-        name: "sp-dashboard",
+        name: "{{ manage.sp_dashboard_user }}",
```
quartje: Unfortunately, this will lead to a recursive loop problem.  Better use a separate variable, like sp_dashboard_manage_user or something

PR #156, opened 2017-11-22
Add checks to prevent undefined variable errors

PeterMosmans:
Before checking whether an attribute of a variable is defined, first check
whether the variable itself is defined.
This prevents errors if the variables themselves aren't (properly) defined while
executing the playbook and roles.


baszoetekouw:
Hi Peter!  Thanks for this patch!

I am wondering though: why is this change necessary?  In the context of OpenConext, we kind of regard the current behaviour as a feature:  running the playbook without having variables like `apache_app_listen_address` defined _should_ result in an error because it means your groups_vars are not set up correctly.

I would agree with you that making such checks explicit would be nicer, but it would also be a lot of work for very little gain.

Could you please explain why you would like to make these changes?  Are you trying to use the roles outside the scope of OpenConext, perhaps?

PR #157, opened 2017-11-22
Postfix part of the pseudonymous email functionality

baszoetekouw:
These is the setup for the mailserver (postfix) on an Openconext machine to forward and rewrite emails for a pseudonymous email domain.  

The corresponding code in AA-server, to rewrite the addresses during a SAML session, was implemented here: https://github.com/OpenConext/OpenConext-attribute-aggregation/commit/b51acea570470defcfdd8a5afa727b4e62df3f05

### DIFF
--- a/roles/common/templates/main.cf.j2
+++ b/roles/common/templates/main.cf.j2
@@ -681,3 +681,9 @@ readme_directory = /usr/share/doc/postfix-2.10.1/README_FILES
 
 smtp_tls_loglevel = 1
 smtp_tls_security_level = may
+
+{% if handle_pseudo_email %}
+virtual_alias_domains = {{attribute_aggregation_pseudo_mail_postfix}}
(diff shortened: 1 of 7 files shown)


roles/common/templates/main.cf.j2
```diff
@@ -681,3 +681,11 @@ readme_directory = /usr/share/doc/postfix-2.10.1/README_FILES
 
 smtp_tls_loglevel = 1
 smtp_tls_security_level = may
+
+{% if handle_pseudo_email %}
+canonical_maps = mysql:/etc/postfix/mysql-pseudo.cf
+remote_header_rewrite_domain = anomail.conext.surfnetlabs.nl
```
quartje: This hostname is SURFnet specific. Changing it into a variable makes it usable outside the SURFnet context as well.

baszoetekouw: Uh woops, that's a stupid oversight, as the variable is already defined.  I'll fix this.

PR #158, opened 2017-11-22
Make disabling of SELinux reentrant

baszoetekouw:


quartje:
Setting selinux to permissive does take quite some effort ... Thanks!

PR #159, opened 2017-11-24
Remove duplicate configuration for SSL

khorben:
Some configuration files for Apache define SSL certificates twice. This
is not supported and breaks the deployment, as Apache will not start.

### DIFF
--- a/roles/metadata/templates/metadata.conf.j2
+++ b/roles/metadata/templates/metadata.conf.j2
@@ -11,13 +11,6 @@ Listen {{ apache_app_listen_address.metadata }}:{{ loadbalancing.metadata.port }
           Require all granted
     </Directory>
 
-{% if apache_app_listen_address.all is defined %}
-    SSLEngine on
-    SSLCertificateFile      {{ tls.cert_path }}/{{ tls_star_cert }}
-    SSLCertificateKeyFile   {{ tls.cert_private_path }}/{{ tls_star_cert_key }}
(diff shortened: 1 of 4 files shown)


quartje:
This double configuration is there for a reason, there are two conditions in the Apache configuration files for using TLS:

1)    {% if haproxy_backend_tls %}

This TLS configuration is used when the application is behind a  haproxy loadbalancer, and needs to use TLS to encrypt the traffic between the loadbalancer and Apache. 
It can be enabled by setting this variable to true in the group_vars

2    {% if apache_app_listen_address.all is defined %}
 
This variable is defined when no loadbalancer is used and Apache is directly connected to the internet. See chapter minimal installation feature on https://github.com/OpenConext/OpenConext-deploy/wiki/Installation-steps-to-deploy-OpenConext-on-a-single-system-other-than-the-Vagrant-VM-centOS7 for more info.

Bottom line is that both variables should not be enabled at the same time. 

Regards,
Bart

khorben:
Please read the patch again, the sections `{% if apache_app_listen_address.all is defined %}` are duplicated. I did not touch anything about HAProxy.

khorben:
In the output from `git diff` it is not obvious, but the same sections (plus `include "ssl_backend.conf"`) are found further down in each file modified, within the same virtual host or section.

quartje:
You're absolutely right, I misread the patch. Thanks for fixing this!

PR #163, opened 2018-02-01
Set profile.baseUrl during provisioning

MKodde:
In OpenConext/engineblock [PR#472](https://github.com/OpenConext/OpenConext-engineblock/pull/472) the profile.baseUrl was added. This
commit ensures that the parameter is added during provisioning of
EngineBlock.


thijskh:
Not sure if we can make EB dependent on defaults of profile?
I would add it to EB's defaults with value `https://profile.{{ base_domain }}`.

PR #165, opened 2018-02-05
Migrated from Service-Registry to Manage in the VM

oharsta:
SR is no longer installed and the sql scripts are also not inserted.
Instead the metadata in imported in mongodb and the push from Manage to
EngineBlock is kicked off using the uri module in ansible.

Smooth migration...

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,11 @@ These files are licensed under version 2.0 of the Apache License, as described i
 To provision the VM use the following (password is vagrant and sudo password is <enter>
 
 ```
-ansible-playbook -u vagrant -i inventory/vm -K selfservice.yml -k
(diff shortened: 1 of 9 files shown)


oharsta:
Service Registry database is still being used by other applications. We first need to migrate those dependencies. For now we leave service registry and janus in the VM.

MKodde:
Thanks that makes sense, and your commit history should have answered my question 👍

quartje:
In the SR provisioning role the feature minimal_install was present, which variable presence will only install EB, profile, SR and manage. It would be nice to have that feauture in this provisioning role as well. 
Secondly, some variables are hardcoded again (the hostnames for instance, and the {{ instance_name }} and {{ support_email }}. When deploying from a template that'd be nice to have as well.

PR #166, opened 2018-02-19
Remove dbpatch migration from EB tasks

jorissteyn:
Database schema changes are no longer managed using 'dbpatch' in EB5.7
and higher.

See: https://www.pivotaltracker.com/story/show/148324779


jorissteyn:
https://github.com/OpenConext/OpenConext-engineblock/pull/493

jorissteyn:
@quartje the build fails because travis tries to execute all migrations since 2012. The change I'm proposing will intentionally fix the upgrade path from EB<5.0. Can we update ./roles/vm_only_provision_eb_sr/templates/engineblock.sql.j2? How?

thijskh:
Travis build should be switched to the new Manage way of workung to bootstrap the system.

thijskh:
@oharsta can you help with that?

jorissteyn:
The task `roles/engineblock/tasks/main.yml`  runs this command after importing the SQL dump:

    command: "app/console doctrine:migrations:migrate -n --env={{ engine_apache_symfony_environment }}"

But starting from EB 5.7 the complete schema (without data) can be created using:

    command: "app/console doctrine:schema:create -n --env={{ engine_apache_symfony_environment }}"

quartje:
Is there a app/console command that provides a reliable way of detecting whether the database is present and provisioned? If so we can add a "app/console doctrine:schema:create -n --env={{ engine_apache_symfony_environment }}" command to the engineblock role, and remove loading the engineblock sql file from the role vm_only_provision_eb_sr

MKodde:
I guess you could use the output of `app/console doctrine:migrations:status` to determine if the database is present and provisioned.

Example output of this console command:
```bash
> app/console doctrine:migrations:status

 == Configuration

    >> Name:                                               OpenConext EngineBlock Migrations
    >> Database Driver:                                    pdo_mysql
    >> Database Name:                                      eb
    >> Configuration Source:                               manually configured
    >> Version Table Name:                                 migration_versions
    >> Version Column Name:                                version
    >> Migrations Namespace:                               OpenConext\EngineBlock\Doctrine\Migrations
    >> Migrations Directory:                               /opt/openconext/OpenConext-engineblock/app/../database/DoctrineMigrations
    >> Previous Version:                                   2018-02-19 17:15:31 (20180219171531)
    >> Current Version:                                    2018-02-20 08:05:49 (20180220080549)
    >> Next Version:                                       Already at latest version
    >> Latest Version:                                     2018-02-20 08:05:49 (20180220080549)
    >> Executed Migrations:                                14
    >> Executed Unavailable Migrations:                    1
    >> Available Migrations:                               13
    >> New Migrations:                                     0
Connection to 127.0.0.1 closed.
```

Or when no database is present

quartje:
That seems like the right approach, it leads to a much cleaner way of bootstrapping the VM. 
I tried to test this, but that leads to some problems:

I tried to provision a database from scratch using EB 5.6.2 with this command:
`command: "app/console doctrine:schema:create -n --env={{ engine_apache_symfony_environment }}"`
That creates a database, but consent is missing from that database. 
If I execute afterwards: 
`app/console doctrine:migrations:migrate -n --env={{ engine_apache_symfony_environment }}`
it fails with because it tries to execute changes that are already in the database.

MKodde:
Could it be you are using an older version of EngineBlock while running the `schema:create` statement?. Quite recently @jorissteyn added Doctrine entities to be able to create the Consent table.

MKodde:
What's the state of this PR? I see some recent changes from @quartje.

quartje:
The last two commits should do the trick and let EB provision a database when none is present. This PR can be merged when EB 5.7 is stable and released in OpenConext-deploy

thijskh:
Rebased on and resolved conflicts with current `master`. Still failing though…

PR #167, opened 2018-02-20
Add EngineBlock test database to provisioning

MKodde:
When building a development environment, we also need a test database to be able to run functional tests.

This change adds the eb_test database and provisions a test user. The test database config is also added to the engineblock.ini template file.

@quartje or @thijskh , some questions: 
* I'm not realy sure I also needed to add the eb-test database config to the docker group_vars. Is this a requirement? 
* Are functional tests run in the docker environment? 
* How do I prevent the test database from being provisioned in production builds?

### DIFF
--- a/roles/engineblock/tasks/main.yml
+++ b/roles/engineblock/tasks/main.yml
@@ -66,6 +66,14 @@
       chdir: "{{ openconext_releases_dir }}/OpenConext-engineblock"
   changed_when: False
 
+- name: Create EngineBlock test database
+  mysql_db: name={{ engine_test_database_name }} state=present
+  when: develop
+
+- name: Create EngineBlock test database user
+  mysql_user: name={{ engine_test_database_user }} host={{ engine_test_database_host }} password={{ engine_test_database_password }} priv={{ engine_test_database_name }}.*:ALL state=present
+  when: develop
+
 # Any scripts should be placed below this
 - name: Run EngineBlock DbPatch migrations
   command: ./bin/migrate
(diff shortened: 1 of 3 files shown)


MKodde:
After being provided with feedback (through Slack) I've made the following changes to this PR:
1. Docker VM is not provisioned with a test database. In fact only the dev vm is provided with an `eb_test` database schema and user.
2. Changes to the `template.yml` have been revoked.
3. The test database is only provisioned when provisioning in 'dev' mode.

roles/engineblock/tasks/main.yml
```diff
@@ -66,6 +66,14 @@
       chdir: "{{ openconext_releases_dir }}/OpenConext-engineblock"
   changed_when: False
 
+- name: Create EngineBlock test database
+  mysql_db: name={{ engine_test_database_name }} state=present
+  tags: mysql
```
quartje: please add a "when: develop" here so it will only be provisioned on the dev vm

roles/engineblock/tasks/main.yml
```diff
@@ -66,6 +66,14 @@
...
+
+- name: Create EngineBlock test database user
+  mysql_user: name={{ engine_test_database_user }} host={{ engine_test_database_host }} password={{ engine_test_database_password }} priv={{ engine_test_database_name }}.*:ALL state=present
+  tags: mysql
```
quartje: Please add a "when: develop" here as well

MKodde: Ah yes, missed that one. Thanks for the feedback. I've replaced the mysql tag with the when: develop statement.

PR #174, opened 2018-05-22
Feature/haproxy enhancements

quartje:
Two features are in this PR:

1) Split up the configuration file in multiple parts
2) In a multi instance setup, and the ability to have an extra instance that can manage http redirects


thijskh:
Yes, this looks good to me.
Do wonder if there's a lot of value to keep the single instance setup vs the duplication it gives (and therefore also means any bugs that show only in the multi-instance are found much later).

thijskh:
The redirects feature is not configured anywhere at the moment, right?

quartje:
The complexity remains indeed. A refactor could be done, but making it less complex and still have the ability to allow blocking access to restricted application can be challenging.

PR #186, opened 2018-08-09
Remove Janus references from eb.ini file

MKodde:
This is part of the cleanup that was performed in
OpenConext-engineblock. When building a new EngineBlock development env,
EB will fail on invalid configuration that is provided in the two ini
files. The application.ini file (override) was updated in the EB repo.
The one written to /etc/openconext was yet to be updated.

And that is taken care of in this PR.

See https://github.com/OpenConext/OpenConext-engineblock/pull/581

### DIFF
--- a/environments/template/secrets/skeleton.yml
+++ b/environments/template/secrets/skeleton.yml
@@ -21,7 +21,7 @@ mongo_passwords:
 mongo_admin_password: secret 
 
 engine_janus_secret: secret
-engine_api_janus_password: secret
+engine_api_metadata_push_password: secret
 engine_api_profile_password: secret
 engine_api_deprovision_password: secret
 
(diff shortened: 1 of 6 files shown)


environments/template/secrets/skeleton.yml
```diff
@@ -21,7 +21,7 @@ mongo_passwords:
 mongo_admin_password: secret 
 
 engine_janus_secret: secret
```
pablothedude: This one is valid?

pablothedude: `engine_janus_secret` seems to be missing on a few other places too

MKodde: I've left this out of scope even tho it is used in the `.ini` file. In my opinion, the complete removal of Janus, including the install and uninstall tasks should be outside of the scope of this PR.

PR #196, opened 2019-01-18
Feature/shibboleth session in db

quartje:


tvdijen:
I use memcache to store shib sessions in my environments.. I can share my config if anyone's interested

quartje:
@tvdijen : I'd certainly be interested in your configuration. For now we're using Galera to store them since we have that and it works fine, but that could change in the future of course.

PR #198, opened 2019-03-27
Add selenium to EB task to allow browser testing

pablothedude:
This will make sure selenium exists and a service is added to
allow browser based testing with Selenium.

### DIFF
--- a/roles/engineblock/tasks/test.yml
+++ b/roles/engineblock/tasks/test.yml
@@ -1,0 +1,30 @@
+---
+- name: Create install directory
+  file:
+    path: "{{ engineblock_selenium_install_dir }}/selenium"
+    state: directory
+    recurse: yes
+
+- name: Install Selenium standalone
+  get_url:
+    url: "http://selenium-release.storage.googleapis.com/{{ engineblock_selenium_version | regex_replace('\\.[0-9]+$', '') }}/selenium-server-standalone-{{ engineblock_selenium_version }}.jar"
+    dest: "{{ engineblock_selenium_install_dir }}/selenium/selenium-server-standalone-{{ engineblock_selenium_version }}.jar"
+    mode: 0755
+
+- name: Install ChromeDriver
+  unarchive:
+    src: "https://chromedriver.storage.googleapis.com/{{ engineblock_chromedriver_version }}/chromedriver_linux64.zip"
+    dest: "{{ engineblock_selenium_install_dir }}/selenium"
+    mode: 0755
+    copy: no
+
+- name: Install Chrome
+  yum:
+    name: https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
+    state: present
+
+- name: Install systemd unit file
+  template:
+    src: "selenium-unit.j2"
+    dest: /etc/systemd/system/selenium.service
+    mode: 0755
--- a/roles/engineblock/templates/selenium-unit.j2
+++ b/roles/engineblock/templates/selenium-unit.j2
@@ -1,0 +1,11 @@
+[Unit]
+Description=Selenium test framework
+After=syslog.target network.target
+
+[Service]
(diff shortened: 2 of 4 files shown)


pablothedude:
@quartje could you thoroughly verify if this PR is in compliance with your standards?

roles/engineblock/tasks/test.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Create install directory
+  file: "path={{ engineblock_selenium_install_dir }}/selenium state=directory recurse=yes"
+
```
quartje: A minor cosmetic nitpick: Could you use the same indentation style as the other tasks?

pablothedude: Thanks, changed the style

pablothedude:
@MKodde could you review this change. In order to test this change you could use:
`./provision vm --tags eb -e develop=true`

roles/engineblock/tasks/test.yml
```diff
@@ -0,0 +1,27 @@
...
+    mode: 0755
+
+- name: Install Chromedriver
+  unarchive:
+    src: "https://chromedriver.storage.googleapis.com/{{ engineblock_chromedriver_version }}/chromedriver_linux64.zip"
+    dest: "{{ engineblock_selenium_install_dir }}/selenium"
+    mode: 0755
+    copy: no
+
+- name: Install Chrome
```
MKodde: The installation of Chrome is now performed twice, as puppeteer also has this as a dependency and will also install it. In my opinion, installing it using OC-Deploy is the better solution. But it might be worthwile to check if puppeteer could be instructed to use this Chrome installation instead of pulling it in by itself.

roles/engineblock/tasks/test.yml
```diff
@@ -0,0 +1,27 @@
...
+- name: Create install directory
+  file: "path={{ engineblock_selenium_install_dir }}/selenium state=directory recurse=yes"
+
+- name: Install Selenium standalone
+  get_url:
+    url: "http://selenium-release.storage.googleapis.com/{{ engineblock_selenium_version | regex_replace('\\.[0-9]+$', '') }}/selenium-server-standalone-{{ engineblock_selenium_version }}.jar"
+    dest: "{{ engineblock_selenium_install_dir }}/selenium/selenium-server-standalone-{{ engineblock_selenium_version }}.jar"
+    mode: 0755
+
+- name: Install Chromedriver
```
MKodde: Nit: ~~Chromedriver~~ ChromeDriver

roles/engineblock/tasks/test.yml
```diff
@@ -0,0 +1,27 @@
...
+    dest: "{{ engineblock_selenium_install_dir }}/selenium"
+    mode: 0755
+    copy: no
+
+- name: Install Chrome
+  yum:
+    name: https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
+    state: present
+
+- name: Install systemd unit file
```
MKodde: Nice solution!

roles/engineblock/templates/selenium-unit.j2
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=selenium test framework
```
MKodde: Nit: uppercase Selenium?

roles/engineblock/tasks/test.yml
```diff
@@ -0,0 +1,30 @@
...
+    src: "https://chromedriver.storage.googleapis.com/{{ engineblock_chromedriver_version }}/chromedriver_linux64.zip"
+    dest: "{{ engineblock_selenium_install_dir }}/selenium"
+    mode: 0755
+    copy: no
+
+- name: Install Chrome
+  yum:
+    name: https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm
+    state: present
+
```
quartje: You could also opt to install chromedriver from the local YUM repositories: It will add chromium as a dependency

pablothedude: I've considered using yum but chrome is also used for frontend testing and using the latest and greatest would prevent much incompatibility issues there.

PR #199, opened 2019-04-19
Add configuration settings for memory limit for push

pablothedude:
The metadata memory limit needs to be configurable. Therefore we
need to add this setting to deploy.

### DIFF
--- a/roles/engineblock/defaults/main.yml
+++ b/roles/engineblock/defaults/main.yml
@@ -80,6 +80,9 @@ engine_fpm_memory: 128M
 # When using vagrant for provisioning, it's not possible to put something in the inventory dir
 engine_crt_not_in_inventory: false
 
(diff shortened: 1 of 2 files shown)


roles/engineblock/defaults/main.yml
```diff
@@ -80,6 +80,9 @@ engine_fpm_memory: 128M
 # When using vagrant for provisioning, it's not possible to put something in the inventory dir
 engine_crt_not_in_inventory: false
 
+# The memory limit used for the metadata push
+metadata_push_memory_limit: 256M
+
```
quartje: Could you prefix this variable with engine_ ? It makes it easier to recognise if you override it somewhere

pablothedude: @quartje done

PR #200, opened 2019-04-26
Add configuration option to set EB theme

pablothedude:
An option is added to the EB configuration in order to set the EB theme
and to make it configurable during deployment.


pablothedude:
The theme doesn;t have to be configurable

PR #201, opened 2019-04-26
added option for redirect ocsp cron log

renoutg:
Added option for redirection of oscp cron log. example value for ocsp_log_redirect
`ocsp_log_redirect: '&> /var/log/ocsp.log'`


quartje:
Looking good!

PR #203, opened 2019-04-29
Add possibility to set feedback background images

pablothedude:
We need to be able to set a custom background image for the feedback
page. This should be possible now.

### DIFF
--- a/roles/engineblock/tasks/main.yml
+++ b/roles/engineblock/tasks/main.yml
@@ -64,6 +64,30 @@
   when:
     customlogo.stat.exists
 
+- name: Check if we have a custom background back image for the feedback page
+  local_action: stat path="{{ inventory_dir }}/files/eb/background-back.svg"
+  register: eb_customfeedbackbackground
+  become: false
+
+- name: Install environment specific background back image
+  copy:
+    src: "{{ inventory_dir }}/files/eb/background-back.svg"
+    dest: "{{ openconext_releases_dir }}/OpenConext-engineblock/web/images/"
+  when:
+    eb_customfeedbackbackground.stat.exists
+
+- name: Check if we have a custom background front image for the feedback page
+  local_action: stat path="{{ inventory_dir }}/files/eb/background-front.svg"
+  register: eb_customfeedbackforeground
+  become: false
+
+- name: Install environment specific background front image
+  copy:
+    src: "{{ inventory_dir }}/files/eb/background-front.svg"
+    dest: "{{ openconext_releases_dir }}/OpenConext-engineblock/web/images/"
+  when:
+    eb_customfeedbackforeground.stat.exists
+
 - name: Install Apache vhost
   template:
     src: "{{ item }}.j2"


pablothedude:
@quartje is this a workable solution to overwrite the feedbackground images? I've added the two images you could use for the private environment in the [pivotal issue](https://www.pivotaltracker.com/story/show/165240377/comments/202086554).

roles/engineblock/tasks/main.yml
```diff
@@ -64,6 +64,30 @@
...
     customlogo.stat.exists
 
+- name: Check if we have a custom background back image for the feedback page
+  local_action: stat path="{{ inventory_dir }}/files/background-back.svg"
+  register: customfeedbackbackground
+  become: false
+
+- name: Install environment specific background back image
+  copy:
+    src: "{{ inventory_dir }}/files/background-back.svg"
```
quartje: could you put in in the subdirectory {{ inventory_dir }}/files/eb ? Makes it easier to differentiate between files for different applications

roles/engineblock/tasks/main.yml
```diff
@@ -64,6 +64,30 @@
   when:
     customlogo.stat.exists
 
+- name: Check if we have a custom background back image for the feedback page
+  local_action: stat path="{{ inventory_dir }}/files/background-back.svg"
+  register: customfeedbackbackground
```
quartje: Could you prefix customfeedbackground with eb_ ?

roles/engineblock/tasks/main.yml
pablothedude: :+1:

roles/engineblock/tasks/main.yml
pablothedude: :+1:

roles/engineblock/tasks/main.yml
```diff
@@ -64,6 +64,30 @@
   when:
     customlogo.stat.exists
 
+- name: Check if we have a custom background back image for the feedback page
+  local_action: stat path="{{ inventory_dir }}/files/eb/background-back.svg"
+  register: eb_customfeedbackbackground
```
MKodde: `customfeedbackbackground` seems odd. I'm not fully familiar with the Ansible naming conventions, but I suspect you are allowed to use underscores to make the register more readable.

`eb_custom_feedback_background`

roles/engineblock/tasks/main.yml
```diff
@@ -64,6 +64,30 @@
...
+  become: false
+
+- name: Install environment specific background back image
```
MKodde: Proposal to improve: 

> Install environment specific background back image

> Set the environment specific error feedback page background image. (background)

The same suggestion goes for the next taskname, but instead of `(background)` set `(foreground)`

roles/engineblock/tasks/main.yml
pablothedude: It was shamefully pasted from `customlogo` so I thought that was the used convention.

roles/engineblock/tasks/main.yml
pablothedude: It was shamefully pasted from the tasks above and I think it will be sufficient as-is.

PR #209, opened 2019-05-22
OIDCNG well known configuration

oharsta:


roles/oidcng/templates/openid-configuration.json.j2
```diff
@@ -0,0 +1,88 @@
+{
+  "issuer": "https://org.openconext.oidc.ng",
```
quartje: Shouldn't this point to https://oidcng.{{ base_domain }} ?

roles/oidcng/templates/openid-configuration.json.j2
```diff
@@ -0,0 +1,88 @@
...
+    "uids"
+  ],
+  "claims_parameter_supported": true,
+  "acr_values_supported": [
+    "http://{{ env }}.surfconext.nl/assurance/loa1",
+    "http://{{ env }}.surfconext.nl/assurance/loa2",
+    "http://{{ env }}.surfconext.nl/assurance/loa3",
+    "http://{{ env }}.surfconext.nl/assurance/sfo-level2",
+    "http://{{ env }}.surfconext.nl/assurance/sfo-level3"
+  ],
```
quartje: too much SURFconextisms here. Can we make this switchable, and the values configurable?

roles/oidcng/tasks/main.yml
```diff
@@ -63,6 +63,13 @@
   notify:
     - "restart oidcng"
 
+- name: copy oidcng well-known discovery configuration
+  template:
+    src: "openid-configuration.json.j2"
+    dest: "{{ oidcng_dir }}/openid-configuration.json"
+  notify:
+    - "restart oidcng"
```
quartje: Do we need to restart or would a reload be sufficient?

oharsta: Restart and reload are the same for a Spring boot application

roles/oidcng/templates/openid-configuration.json.j2
oharsta: https://github.com/OpenConext/OpenConext-deploy/commit/93e49d1106123949a163a59de7b4f47d9a61c031

roles/oidcng/templates/openid-configuration.json.j2
oharsta: Yep. Changed this.

PR #210, opened 2019-05-23
Feature/oidc10rp allowed resource servers

oharsta:


roles/oidcng/tasks/main.yml
```diff
@@ -63,6 +63,13 @@
   notify:
     - "restart oidcng"
 
+- name: copy oidcng well-known discovery configuration
+  template:
+    src: "openid-configuration.json.j2"
+    dest: "{{ oidcng_dir }}/openid-configuration.json"
+  notify:
+    - "restart oidcng"
+
```
quartje: Can this be a reload?

roles/oidcng/templates/openid-configuration.json.j2
```diff
@@ -0,0 +1,88 @@
...
+    "http://{{ env }}.surfconext.nl/assurance/loa2",
+    "http://{{ env }}.surfconext.nl/assurance/loa3",
+    "http://{{ env }}.surfconext.nl/assurance/sfo-level2",
+    "http://{{ env }}.surfconext.nl/assurance/sfo-level3"
+  ],
+  "code_challenge_methods_supported": [
+    "plain",
+    "S256"
+  ]
+}
```
quartje: I reviewed this change in the other PR

roles/oidcng/tasks/main.yml
oharsta: Reload and restart are the same for Spring boot app

roles/oidcng/templates/openid-configuration.json.j2
oharsta: That's correct. I needed to merge feature branches otherwise the deploy would fail

PR #215, opened 2019-05-30
Feature/oidc playground

oharsta:


roles/oidc-playground-client/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
+---
+- name: create apache folder
+  file:
+    path: "/var/www/oidc-playground"
+    owner: apache
```
quartje: owner should be root, we don't want to make the web server able to write in this directory

roles/oidc-playground-server/handlers/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  systemd:
+    daemon_reload: yes
+
+- name: restart oidc-playground
+  service:
+    name: oidc-playground
+    state: restarted
+
+- name: update_ca_trust
+  command: update-ca-trust
```
quartje: I don't see this handler used in this role, so it can be removed here

oharsta: Done

roles/oidc-playground-client/tasks/main.yml
oharsta: Done

PR #222, opened 2019-06-13
Install PHP 7.2 alongside 5.6, 7.3

MKodde:
In order to test PHP 7.2 for compatibility issues, PHP 7.2 with the PHP-FPM  and extensions are installed.


thijskh:
Continued in #233

PR #229, opened 2019-06-21
Feature/haproxy refactor

quartje:
This PR will eleminate the haproxy multi instance option, which reduces the complexity of the Haproxy role significantly.
It will also install haproxy 1.8 from softwarecollections.org, in stead of relying on a RPM repo hosted by SURFnet. If you run a single instance Haproxy installation, the upgrade will happen automatically.


roles/keepalived/templates/keepalived_loadbalancer.conf.j2
```diff
@@ -2,215 +2,31 @@
...
-   advert_int 5
-   virtual_ipaddress {
-       {% if engine_ipv6 is defined %}
-       {{ engine_ipv6 }}
-       {% endif %}
-       {% if bind_ipv6 is defined %}
-       {{ bind_ipv6 }}
+       {{ haproxy_sni_ip.ipv4 }}
+       {% if haproxy_sni_ip_restricted.ipv4 is defined %}
+       {{ haproxy_sni_ip.ipv4 }}
```
thijskh: haproxy_sni_ip_restricted.ipv4

roles/haproxy/files/haproxy_systemd_redirects
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=HAProxy Load Balancer Redirects instance
+After=network.target
+
+[Unit]
+Description=HAProxy Load Balancer
+After=network.target
```
thijskh: This section is duplicate

environments/template/group_vars/template.yml
```diff
@@ -623,8 +623,13 @@ haproxy_applications:
...
 
-haproxy_sni_ip: "*"
-haproxy_sni_ipv6: "::1"
+haproxy_sni_ip:
+  ipv4: "*"
+  ipv6:  "::1"
+  certs: 
+     - name: star
+       key_content: "{{ https_star_private_key }}" 
+       crt_name: star.{{ base_domain }}.pem
```
thijskh: The template and vm environments do not define a restricted ip. This means that those parts of the haproxy config are not run in tests and not on simpler installations, so get less testing exposure. Would there be a way to also define a haproxy restricted ip and run the restricted apps on that?

quartje: Yes, that's a good idea to test as well. I've added a second ip: https://github.com/OpenConext/OpenConext-deploy/pull/229/commits/f6b5c5b6ffbde0f178edbcdd6e53831349d3a8c7

PR #230, opened 2019-06-26
Added required input parameters for SURFmarket entitlements

oharsta:


thijskh:
Why is `urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified` added and not `urn:mace:dir:attribute-def:uid`?

oharsta:
See https://github.com/OpenConext/OpenConext-deploy/commit/0dbbe2ee8339ad9102e090134abba1eb02abfd7a

PR #233, opened 2019-07-08
Install PHP 7.2 instead of 7.3

thijskh:


roles/php/files/remi-php72.repo
```diff
@@ -0,0 +1,34 @@
...
+enabled=0
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-remi
+
+[remi-php72-test-debuginfo]
+name=Remi's PHP 7.2 test RPM repository for Enterprise Linux 7 - $basearch - debuginfo
+baseurl=http://rpms.remirepo.net/enterprise/7/debug-test72/$basearch/
+enabled=0
+gpgcheck=1
+gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-remi
```
quartje: Maybe we can leave out the test and debuginfo repos. That would save some time during updates.

thijskh: They are not enabled

roles/php/tasks/main.yml
```diff
@@ -34,30 +34,30 @@
       - composer
     state: present
 
-- name: check for presence of the remi 73 repository file
+- name: check for presence of the remi 72 repository file
   stat:
-    path: "/etc/yum.repos.d/remi-php73.repo"
-  register: remi_php73_config
+    path: "/etc/yum.repos.d/remi-php72.repo"
+  register: remi_php72_config
```
quartje: I don't think we need to register the presence of this file. The next copy task will take care of that automatically

MKodde:
In order to get EB to chooch, I needed some additional bells and whistles. Chief amongst which is the apc-bc PHP 7.2 extension. This is required to satisfy the version of Doctrine that we use. So this extension can probably be skipped if we upgrade to a newer Doctrine version.

See steps 8 and 9 of https://github.com/OpenConext/OpenConext-engineblock/pull/713#issue-287892008

As for the other steps. OC deploy can easily be configured to take care of those additional configuration steps.

MKodde:
And an additional todo is that the xdebug extension is not configured correctly (does not work). I have not yet fixed this in my box, but will come back with more details if this becomes a problem.

thijskh:
I have merged this to master now. Anything that needs to be improved can be done separately.

PR #243, opened 2019-08-02
Update chromedriver for EB tests

pablothedude:
This will make sure the latest chromedriver is used. This stopped
working in Travis ans is therefore also fixed here.


pablothedude:
@quartje would you be so kind to review this PR? This was fixed in EB for Travis but I would like the changes also here so if someone starts developing this should work out of the box.

https://github.com/OpenConext/OpenConext-engineblock/pull/728

PR #244, opened 2019-08-13
PHP 7.3 cleanup and development change

MKodde:
1. PHP 7.3 is stopped from being installed
2. The app_dev.php.dist from the OpenConext-engineblock project is used to put in the EB web folder instead of the locally tracked app_dev.php file that was previously used.


thijskh:
See #233 for my proposal that will result in both 5.6 and 7.2 installed.

MKodde:
I'm closing this PR in favor of #233 which makes this PR redundant, and adds other cool stuff.

PR #252, opened 2019-10-22
Remove: serviceregistry/janus/metadata-exporter/grouper legacy

thijskh:


quartje:
The move to php 7.2 seems like a good moment to ditch support for Janus and other legacy apps.
There's one point where serviceregistry is still mentioned: https://github.com/OpenConext/OpenConext-deploy/blob/master/group_vars/all.yml#L27. 
Changing it there would require both Manage _and_ eb to be deployed at the same time. 
Everything else is looking good

thijskh:
Yes, that's why I kept it - it's only a name so not worth the trouble to change in my opinion.

PR #253, opened 2019-11-15
Feature/eb 6.0 php7.2

thijskh:


roles/engineblock/tasks/install-branch.yml
```diff
@@ -37,3 +37,4 @@
   notify:
     - "restart httpd"
     - "restart php-fpm"
```
quartje: We don't need "restart php-fpm" anymore. The app/console commands are also php72 only

thijskh:
I was planning to remove that everywhere when we also remove installing php5 from the playbooks. Not sure which way is better.

quartje:
It doesn't hurt either ...

MKodde:
**Observation:**
this PR adds support for the changes introduced in: https://github.com/OpenConext/OpenConext-engineblock/pull/800 :+1:

PR #257, opened 2019-12-17
removed and changed legacy brew commands

Timdebruijn:
In newer versions of brew it's not needed to install cask, this is already included in brew.
"brew tap caskroom/versions" will give the following error:
caskroom/versions was moved. Tap homebrew/cask-versions instead.
So I changed that to "brew tap homebrew/cask-versions"


quartje:
Thanks Tim!

PR #258, opened 2019-12-19
Feature/haproxy send samesite none cookie attribute

quartje:
This PR tries to work around the new default setting of Chrome for the samesite cookie attribute. The new default (samesite=origin) makes that in cross site POSTS the cookie is no longer included in the POST. This creates issues for our loadbalancing and cookies for SAML sessions.  See https://www.chromium.org/updates/same-site for more info.

In this PR, Haproxy rewrites _all_ cookies including the loadbalancer cookies, adding "samesite=none". Since there are some browser incompabilities (https://www.chromium.org/updates/same-site/incompatible-clients) ACLs are in place to recognise those, and override the rewrite behaviour

### DIFF
--- a/roles/haproxy/tasks/main.yml
+++ b/roles/haproxy/tasks/main.yml
@@ -139,6 +139,14 @@
   notify:
     - "reload haproxy"
 
+- name: Place file with list of browsers that do not accept samesite=none cookies
+  copy:
+    src: nosamesitebrowsers.lst
+    dest: /etc/haproxy/nosamesitebrowsers.lst
+    owner: root
+    group: root
+    mode: 0744
+
 - name: Run hapos-upd for initial OCSP responses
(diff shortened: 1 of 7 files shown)


quartje:
I think this https://github.com/OpenConext/OpenConext-deploy/pull/258/commits/ff60c041fa82194997f2e6db21bc33de7bf7f65a commit simplifies it as you suggested. It'd be even nicer if we could somehow combine the two frontends (restricted and unrestricted) into one. For now I've added a file that loads the regular expressions for browser detection

roles/haproxy/tasks/main.yml
```diff
@@ -139,6 +139,14 @@
   notify:
     - "reload haproxy"
 
+- name: Place file with list of browsers that do not accept samesite=none cookies
+  copy:
+    src: nosamesitebrowsers.lst
+    dest: /etc/haproxy/nosamesitebrowsers.lst
+    owner: root
+    group: root
+    mode: 0744
```
thijskh: 644?

PR #264, opened 2020-02-03
haproxy: in frontend, also log ciphers used for connection

thijskh:


quartje:
As discussed, I rewrote it a bit so we can extract the information more easy in Logstash. As a bonus, we log the http version now as well

PR #266, opened 2020-02-19
Feature/even more php fixes

quartje:
Uninstall php56 completely, and manually install the remi repo file. Composer is eleminated as well.


roles/php/tasks/main.yml
```diff
@@ -1,23 +1,24 @@
-# Install the "REMI" repo which contain newer php packages that override the default
-# that come with the distro (CentOS7: 5.4; CentOS6: 5.3)
-- name: check for presence of the remi repository file
-  stat:
-    path: "/etc/yum.repos.d/remi.repo"
```
thijskh: Is it now certain that this file is gone?

roles/php/tasks/main.yml
```diff
@@ -30,9 +31,15 @@
       - php72-php-soap
       - php72-php-xml
       - php72-php-gd
-      - composer
     state: present
 
+- name: Remove all php56 packages and composer
+  yum:
```
thijskh: I put removal at the end so the install is not needlessly broken while executing the task. Remove 5.6 when 7.2 is fully configured

roles/php/tasks/main.yml
quartje: yes, it's in the remi-release package, which is being removed. Yum saves a copy in /etc/yum.repos.d/remi.repo.rpmsave

roles/php/tasks/main.yml
quartje: I've added a commit to move it to the end

PR #267, opened 2020-03-25
Move enginblock.ini to parameters.yml

pablothedude:
In order top get rid of the duplicate configuration options from
Symfony (parameters.yml) and the legacy (merged ini) both are merged
into a single parameters.yml.

Also a minor change was done in the prep-dev-env script to be php72
 compatible also in development mode.

https://github.com/OpenConext/OpenConext-engineblock/pull/838
https://www.pivotaltracker.com/story/show/155703980


pablothedude:
In EB  error handling was added to the MakeRelease script and this resulted in a failure of the tests when EB was installed from branch because of missing dependencies. I've updated the install-branch task to install the required dependencies.

https://github.com/OpenConext/OpenConext-engineblock/blob/feature/remove-ini-configuration/bin/makeRelease.sh

Currently only the idempotent fix is failing and I've consulted @thijskh about this and that was acceptable for now.

roles/engineblock/templates/parameters.yml.j2
```diff
@@ -1,4 +1,115 @@
...
+        - http
+        - https
+    php_settings:
+        memory_limit: 128M
+        display_errors: '{{ php_display_errors }}'
+        error_reporting: '6135'
+        date.timezone: {{ timezone }}
+        sendmail_from: 'OpenConext EngineBlock <openconext-engineblock@openconext.org>'
+    api.users.metadataPush.username: {{ engine_api_metadata_push_user }}
+    api.users.metadataPush.password: {{ engine_api_metadata_push_password | replace("%","%%") }}
```
thijskh, line 25: Why is this replace necessary? Can't you just quote the parameters (like sendmail_from above)?

roles/engineblock/templates/parameters.yml.j2
```diff
@@ -1,4 +1,115 @@
...
+    forbidden_signature_methods: {  }
+    allowed_acs_location_schemes:
+        - http
+        - https
+    php_settings:
+        memory_limit: 128M
+        display_errors: '{{ php_display_errors }}'
+        error_reporting: '6135'
+        date.timezone: {{ timezone }}
+        sendmail_from: 'OpenConext EngineBlock <openconext-engineblock@openconext.org>'
```
thijskh, line 23: Really should be `{{ instance_name }} <{{ support_email }}>`

roles/engineblock/tasks/main.yml
```diff
@@ -162,30 +162,13 @@
...
   become: yes
   lineinfile:
     dest: "/etc/environment"
     state: present
     line: "SYMFONY_ENV={{ engine_apache_symfony_environment }}"
 
-# This should always run before any scripts run, it ensures that the env specific
-# configuration is loaded correctly.
-- name: Prepare EngineBlock environment
-  shell: php72 bin/composer/dump-required-ini-params.sh
```
thijskh, line 184: Here we used a composer we apparently already have available. So why is it necessary to install it in this PR?

roles/engineblock/templates/parameters.yml.j2
pablothedude: The problem is that Symfony is using '%' to replace variables. Therefore it needs to get escaped else the character will be removed. This logic was implemented in the dumper script and it was moved here to have the same behaviour.

pablothedude: This is behaviour is of Symfony itself and is on top of the default Yaml behaviour so it wouldn't be suffiecient to just escape Yaml.

roles/engineblock/templates/parameters.yml.j2
pablothedude: Be aware that this seems to change the original behaviour because it didn't seem to be defined in the `engineblock.ini.j2` in deploy and was coming from the default `application.ini` in EB itself. Or does it come from somewhere else and do I missing something here myself?

roles/engineblock/tasks/main.yml
pablothedude: I don't get it, `/bin/composer/` seems to be a folder here?

I've tested the changes in EB with the install from branch task. That task uses the `makeRelease.sh` script in EB and to run that script Composer is needed. I saw Composer was installed earlier on, but I saw efforts of Bart to remove Composer from Deploy and as far as I could see it should only be used when building a tar. I'ver asked Bart why this change was needed and he stated that he didn't like Composer in production and that's something I would totally agree on. So I decided to only added it to the 'install from branch' task were it's needed to build an artifact from scratch and seems to be there to test EB in deploy.

https://github.com/OpenConext/OpenConext-deploy/commit/4f0ef664af88151449e8ecbc9823f92f203d76a2

roles/engineblock/templates/parameters.yml.j2
thijskh: :+1:

roles/engineblock/templates/parameters.yml.j2
thijskh: No just observing that we should be avoiding this kind of hardcoded values. Can indeed be that it was already hardcoded earlier.

roles/engineblock/tasks/main.yml
thijskh: My bad, indeed there's no composer now. Might make sense to move this to the php role in the future if more apps would work in a similar way. But OK for now.

roles/engineblock/templates/parameters.yml.j2
pablothedude: I've changed it as suggested.

thijskh:
Looks good now!
Have to nuance the idempotency thing a bit: it seemed caused by the branch install which is fine. But it should not happen when the commit `Change the EB branch to test changes` is removed.

pablothedude:
@thijskh I don't know if you want the have the enabled languages configurable but I leave it as-is till requested otherwise.
`https://github.com/OpenConext/OpenConext-deploy/pull/267/commits/b0dec7c0da1949237d99344e8b42b2dd5164b78d#diff-ecbc53ffefd8af7e16408f9bd2231180R27`

PR #269, opened 2020-04-20
Feature/haproxy/2.0 upgrade

quartje:
This PR will upgrade Haproxy to the latest 2.0 version. The source for getting the Haproxy package is changed as well, from the Software Collections repo to the IUS repo. The Software Collections repo is not updated frequently enough.

Since the browser detection method does not work anymore after the 2.0 upgrade it has been changed and simplified. It now relies on the Haproxy feature that allows sharing a variable between request and response.


quartje:
I agree that we should look into these features after the upgrade. 
HTTP/2 can be enabled right away again indeed. The version supplied by the ius repo is the one where the HTTP/2 bug (CVE-2020-11100) is fixed (2.0.14)

PR #270, opened 2020-04-30
Create tls selfsignedcerts with ansible

quartje:
In this PR, the creation of the selfsigned certificates is taken away from the BASH scripts, and put into an Ansible role. These certificates can be used with Haproxy, or in the Apache standalone mode. 

It hopefully makes a start with reducing the complexity of setting up a new OpenConext deploy environment. 

This fixes the current issue with the expired VM certificate. It's also expected that the future merging of the VM and Template environment will become easier.


MKodde:
When building a dev vm:

```
$ vagrant destroy
$ ENV=dev vagrant up

TASK [selfsigned_certs : Copy the contents of the certificate to the pki anchor source directory of all VMs] ***
fatal: [192.168.66.99]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'selfsigned_cert_contents' is undefined\n\nThe error appears to be in 'OpenConext-deploy/roles/selfsigned_certs/tasks/main.yml': line 50, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Copy the contents of the certificate to the pki anchor source directory of all VMs\n  ^ here\n"}
```

Version info
Ansible: 2.9.6
Vagrant: 2.2.7
VirtualBox: 6.1.6

quartje:
I always forget that the Vagrant Ansible provisioner works differently from the regular way provisioning is done. You might have more success using that:
```
./provision vm -e engine_apache_symfony_environment=dev -e develop=true 
```
should give better results. It will do a "vagrant up" and run ansible-playbook with the correct parameters. 

I think it's best remove the Ansible provisioner from the Vagrant config alltogether, it has caused issues more than once

roles/selfsigned_certs/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+- name: Get the contents of the certificate
+  slurp:
+    src: "{{ tls.cert_path }}/{{ tls_star_cert }}"
+  register: selfsigned_cert
+  when: "'loadbalancer-vm' in group_names"
+  run_once: true
+
+- name: Put selfsigned cert in a variable
+  set_fact: 
+    selfsigned_cert_contents: "{{ selfsigned_cert.content | b64decode }}"
```
thijskh, line 46: What is base64 decoded and why?

roles/welcome/tasks/main.yml
```diff
@@ -15,11 +15,6 @@
     src: "site/index.html"
     dest: "/var/www/welcome/index.html"
 
-- name: copy the openconext_ca certificate
-  copy:
-    src: "{{ inventory_dir }}/files/certs/{{ tls_ca }}"
-    dest: "/var/www/welcome/openconext_ca.crt"
```
thijskh, line 21: Still linked from index.html

roles/selfsigned_certs/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+    size: 2048
+  when: "'loadbalancer-vm' in group_names"
+
+- name: Generate a CSR
+  openssl_csr:
+    path: /etc/pki/tls/certs/openconext.csr
+    privatekey_path: "{{ tls.cert_private_path }}/{{ tls_star_cert_key }}"
+    common_name: "*.{{ base_domain }}"
+    organization_name: OpenConext
+    basic_constraints: CA:TRUE
```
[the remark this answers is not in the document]
quartje: Yes it is unfortunately. Otherwise the certificate cannot be imported in the certificate store using the method of placing it in /etc/pki/ca-trust/source/anchors/ and running update-ca-trust

roles/selfsigned_certs/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
+---
+- name: Install pip
+  yum:
+    name: "python2-pip"
```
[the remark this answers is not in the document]
quartje: The default python version on CentOS7 is Python2, and Ansible uses that version on the destination host as well. We could change that of course, but that would require another PR.

roles/selfsigned_certs/tasks/main.yml
quartje: The Ansible module "slurp" is used to get the contents of that certificate into a variable, and the set_fact module is then used to make the variable usable on other hosts. The "slurp" module base64 encodes the contents of the file, hence the need to base64 decode it again

roles/welcome/tasks/main.yml
quartje: Fixed in 5a17d7943f08986ec2f3665dee2260687405eced

MKodde:
Test results, some findings:

1. Had to run the provisioning twice to get past the ssh cert signature verification step
2. In order to deploy a modern version of EngineBlock, some changes to the EB roles will be required. Probably already fixed in another PR
3. Composer is not installed on the `apps` server
4. Other than that, the certificates installed without issue! And all tasks passed without error.

PR #271, opened 2020-05-20
Remove unused support urls from EB config

pablothedude:
The support urls are removed in favour of translatable urls
which are configurable through translations.

See: https://github.com/OpenConext/OpenConext-engineblock/pull/851


pablothedude:
The buil di fialing because the change didn't land in EB yet, the build should succeed if the change below lands.
See: OpenConext/OpenConext-engineblock#851

PR #273, opened 2020-05-27
Install EB build tools for EB

pablothedude:
Composer and NPM were not installed in dev mode. This
is because they were initially always installed. But by
removing them from prod they were also missing in dev.


MKodde:
I see no problems with this change. But @quartje might be more able to verify this change! I see the push travis build failed. This seems to be a false positive. Maybe re-run that one?

quartje:
Seems fine to me. The build failures were a known issue. That has been fixed, I'll merge this one.

PR #276, opened 2020-06-22
Feature/haproxy tls ncsc guidelines

quartje:
This PR attempts to bring the TLS settings more in line with the most recent NCSC TLS guidelines (https://english.ncsc.nl/publications/publications/2019/juni/01/it-security-guidelines-for-transport-layer-security-tls). Note that not all guidelines are implemented yet, only the DH exchange parameter file is changed, and the cipher order is changed. 
Disabling TLS 1.0 and TLS 1.1 will follow later.


roles/haproxy/tasks/main.yml
```diff
@@ -105,6 +105,16 @@
     - "reload haproxy"
   when: "haproxy_sni_ip_restricted is defined and 'selfsigned_certs' not in group_names"
 
+- name: Install ffdhe3072 DH parameters
+  file: 
+    src: ffdhe3072.pem
+    dest: /etc/pki/haproxy/ffdhe3072.pem
+    owner: root
+    group: root
+    mode: 0600
```
thijskh, line 114: It's not a secret..

roles/haproxy/tasks/main.yml
```diff
@@ -105,6 +105,16 @@
     - "reload haproxy"
   when: "haproxy_sni_ip_restricted is defined and 'selfsigned_certs' not in group_names"
 
+- name: Install ffdhe3072 DH parameters
+  file:
```
thijskh, line 109: Must be `copy`?

roles/haproxy/tasks/main.yml
quartje: I've change the permissions to 644

PR #284, opened 2020-08-25
Change log path to reflect dev vm path

pablothedude:
In production all is logged to syslog so this wouldn't prove an issue.

See https://github.com/OpenConext/OpenConext-engineblock/pull/857

### DIFF
--- a/roles/engineblock/defaults/main.yml
+++ b/roles/engineblock/defaults/main.yml
@@ -88,7 +88,7 @@ engine_syslog_priority: 7
 #engine_initial_sql: ''
 
 engineblock_symfony_cache_path: "/tmp/engineblock/cache"
-engineblock_symfony_log_path: "/tmp/engineblock/log"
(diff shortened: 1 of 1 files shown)


MKodde (approved):
This should work! @quartje what's your opinion on this change?

quartje:
I don't see the relation with the change in Engineblock: There the cache path is changed, whereas here the logging path is changed.

pablothedude:
@quartje the relation is that for the cache Deploy used `/tmp/engineblock/cache` while EB was using `/tmp/eb_cache/`. For logging in Deploy `/tmp/engineblock/log` was used and EB used `./var/logs`. So the PR's are used to get them both aligned so for caching `/tmp/engineblock/cache` is used and for the logs `./var/logs`.

Also note this is only needed for the development environment.

quartje:
Right, I see the light now :)

quartje (approved):

PR #285, opened 2020-08-31
Feature/java cleanup old releases

quartje:
This will clean up all but the last two versions (the one currently installed, and the previous one) of all Java apps. 
It will keep the filesystems of the Java appservers clean, and make it possible to manually roll back a version by changing the symlink

### DIFF
--- a/roles/account-gui/tasks/main.yml
+++ b/roles/account-gui/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 - name: create apache folder
-  file: path=/var/www/account owner=root group=root mode=2755 state=directory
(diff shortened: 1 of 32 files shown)


quartje:
That's an option to consider. However, the configurations are not uniform over all apps, and sometimes changes are needed (like disabling email logs for instance). All doable I guess, but this gives the largest flexibility. Since the application configuration is also stored in the role as a template I guess we can keep it this way.

quartje:
I've centralised it, and opened another PR for it here: https://github.com/OpenConext/OpenConext-deploy/pull/288
Closing this one

PR #288, opened 2020-09-15
Feature/java apps cleanup and centralise

quartje:
There were multiple installation methods present in tasks/springbootapp. Only install-maven-release.yml is left now.
Some tasks that were present in every java role have been centralised. 
The local installation method for springboot apps has been removed
Some legacy stuff has been removed as well: The removal of init.d style scripts has been removed, some obsolete tags and variables have been removed as well. 
Some tasks remain and will be fixed later:
1) The java heap size is always 512MB now
2) The springbootapp tasks should be put into a role and be included using the "include_role" Ansible module, since the "include" module will be removed in the feature
3) Maybe more could be centralised, like the Apache config and application.yml templates


oharsta:
Looks very good. I see that the override for `springapp_heapsize` is still there, so your comment _The java heap size is always 512MB now_ is luckily not true. We can also remove ` tasks/springbootapp/switch-version.yml`.

quartje:
I managed to screw the Java heapsize setting up with this line:
```
 springapp_min_heapsize if springapp_min_heapsize is defined else springapp_heapsize }} -Xmx{{ springapp_m    ax_heapsize if springapp_max_heapsize is defined else springapp_heapsize
```

As it turns out, role vars are now global :( meaning that if one role in a play has springapp_min_heapsize or springapp_max_heapsize (oidcng in this case), _all_ roles will inherit that var. Anyway, I'll fix it later on by setting specific role overrides for that value.

thijskh (approved):

PR #293, opened 2020-09-29
FIx EB branch deploy + theming

thijskh:


thijskh:
Merging as it improves over current state, branch install is not 100% perfect yet. Theme keeps to be "openconext" for now.

PR #294, opened 2020-09-29
Use the supported_language_codes variable

oharsta:
Now the languages in the patternProperties are hardcoded. Use the
existing supported_language_codes variable to determine the languages.


thijskh (approved):

quartje (approved):
Seems like a flexible solution!

PR #297, opened 2020-10-15
Provision manage entities and migrate authz internal clients to OIDCNG

quartje:
In this PR I've added a role that can provision entities into manage.
The goal is to phase out vm_only_provision_manage_eb which provisions a sql file and that is not very idempotent, nor easy to adapt. All manage entities will then be provisioned from within their own roles. 
Currently, it provisions the three internal client OAuth2 credentials clients : pdp, teams and AA use client credential tokens for accessing Voot. Until now, authz was used as authorization server. Since authz is being phased out, the client credential clients are now migrated to OIDCNG. Deploying pdp, teams and AA will automatically provision these clients, and change the links in the configurations to point to the oidcng token endpoint.


oharsta:
Looks good!

Although there is support for uploading a SP in `roles/manage_provision_entities/tasks/main.yml`, there is no corresponding `roles/manage_provision_entities/templates/saml20_sp.j2`

oharsta (approved):

quartje:
Thanks @oharsta . I'll build in SAML support when there is time, so we can migrate the whole eb_provision_vm_only stuff.

tvdijen:
Is there a guide on how to do this manually through the GUI? I'm a complete oAuth/OIDC noob and I have no idea what I'm support to do to migrate from authz to oidcng

thijskh:
It basically goes as follows:

- You create a new Resource Server for voot in Manage. You configure the entityID (client ID) and secret in voot's application.properties oidcng.* (as described briefly in the readme)
- You create new Relying Parties in Manage for each RP{ currently in authz-admin with the corresponding grants (client_credentials for the platform-internal clients)
- You change the config of the RP's themselves (e.g.: teams, AA, pdp, or external clients) to use the new RP's clientid/secret (if different from the old ones) and the oauth_token_url to that of your oidcng server.

Hope this sets you on the right path, maybe discuss further on Slack if you run into anything more?

tvdijen:
Yes, thank you! I can probably figure this out with the steps above!

PR #299, opened 2020-10-27
Feature/OIDC-NG SAML migration

oharsta:
Required to deploy OIDC-NG 4.0.x releases.


roles/oidcng/templates/application.yml.j2
```diff
@@ -47,38 +50,21 @@ features:
...
+  acs_location: https://{{ oidcng_base_hostname }}/login/saml2/sso/oidcng
+
 idp:
-  alias: engine-{{ env }}
-  # comma separated string of metadata url's
-  metadata_urls: {{ oidcng_idp_metadata_url }}
-  name_id: https://engine.{{ base_domain }}/authentication/idp/metadata
+  entity_id: https://engine.{{ base_domain }}/authentication/idp/metadata
+  sso_location: https://engine.{{ base_domain }}/authentication/idp/single-sign-on
   metadata_signing_certificate_path: file://{{ oidcng_dir }}/SURFconext-metadata-signer.pem
```
quartje, line 60: I assume that this is the SAML assertion signing key? It might make sense to rename this parameter to reflect that

roles/oidcng/templates/application.yml.j2
```diff
@@ -47,38 +50,21 @@ features:
...
+  entity_id: {{ oidcng_saml_sp_entityid }}
+  acs_location: https://{{ oidcng_base_hostname }}/login/saml2/sso/oidcng
+
 idp:
-  alias: engine-{{ env }}
-  # comma separated string of metadata url's
-  metadata_urls: {{ oidcng_idp_metadata_url }}
-  name_id: https://engine.{{ base_domain }}/authentication/idp/metadata
+  entity_id: https://engine.{{ base_domain }}/authentication/idp/metadata
+  sso_location: https://engine.{{ base_domain }}/authentication/idp/single-sign-on
```
quartje, line 59: The SSO location is not always authentication/idp/single-sign-on: It usually include the key slug as well.
Better to make the whole parameter configurable, and let the value "https://engine.{{ base_domain }}/authentication/idp/single-sign-on" be the default in defaults/main.yml

roles/oidcng/templates/application.yml.j2
oharsta: Done

roles/oidcng/templates/application.yml.j2
oharsta: Done

quartje (approved):

PR #301, opened 2020-11-16
Configured "services" in inventory so we can easily deploy the services to the correct hosts

dj-wasabi:

* Added "services" that contain the specific Ansible groups, in which we allow a bit more structure in the `provision.yml` playbook;
* Added new tags that contain the above mentioned "services", but also a tag named `core` that can be used to deploy the core related Ansible roles on the deployment;
* Added a new property `use_selfsigned_certs ` which will install the `selfsigned_cert` role to be executed or not;
* A slightly modified change in the `selfsigned_cert` role to only execute certain tasks on the `loadbalancer` hosts with the new "services" setup;

Changes for inventory & group_vars are both applied to the "vm" as the "template" environments.


provision.yml
```diff
@@ -7,147 +7,181 @@
...
+      tags: ['core', 'base', 'common']
+    - role: rsyslog
+      tags: ['core', 'base', 'rsyslog']
+    - role: iptables
+      when:
+        - iptables_enable | bool
+      tags: ['core', 'base', 'iptables']
+    - role: selfsigned_certs
+      when:
+        - use_selfsigned_certs
```
quartje, line 24: I believe this should be "use_selfsigned_certs | bool "  since the | bool filter is the new default

dj-wasabi: The `| bool` is only needed when the value is written as a string (`"true"`) and not as a bool (`true`). The `use_selfsigned_certs` is written in the group_vars as a boolean (`true`).

provision.yml
```diff
@@ -7,147 +7,181 @@
...
 
-- hosts: loadbalancer_ha
-  gather_facts: true
+- hosts: frontend
   become: true
   roles:
-    - { role: keepalived, tags: ['keepalived'] }
-    - { role: bind, tags: ['bind'] }
+    - role: hosts
+      tags: ['core', 'frontend', 'hosts']
```
quartje, line 31: Maybe create a boolean here as well. The "hosts" role is only for first time deployers where no dns entries are present yet. If we can make this optional it's easier to disable it for hosts where it is not necessary (and could potentially break the installation)

provision.yml
```diff
@@ -7,147 +7,181 @@
...
+    - role: httpd
+      tags: ['core', 'frontend', 'httpd']
+    - role: welcome
+      tags: ['core', 'frontend', 'welcome']
```
quartje, line 35: This role is optional as well. It would be beneficial to get this role behind a feature toggle as well

provision.yml
```diff
@@ -7,147 +7,181 @@
...
+      tags: ['core', 'loadbalancer', 'lb']
+    - role: keepalived
+      when:
+        - inventory_hostname in groups['loadbalancer_ha']
+      tags: ['core', 'loadbalancer', 'keepalived']
+    - role: bind
+      when:
+        - inventory_hostname in groups['loadbalancer_ha']
+      tags: ['core', 'loadbalancer', 'bind']
+
```
quartje, line 53: Same here as for the previous plays: The role "haproxy" is not optional, but the "bind" and "keepalived" are. They are there to deploy the specific roles needed to run OpenConext over several datacenters. We might place these behind feature toggles as well.

provision.yml
dj-wasabi: With first time deployers you don't mean deploying it for the first time and rerun the provision command? As the `hosts` role is idempotent so it won't add things twice.

provision.yml
```diff
@@ -7,147 +7,181 @@
...
+      when:
+        - inventory_hostname in groups['dbcluster_nodes']
+      tags: ['core', 'db_mysql', 'keepalived']
+    - role: galera_create_users
+      when:
+        - inventory_hostname in groups['galera_provision_host']
+      tags: ['core', 'db_mysql', 'galera', 'galera_create_users']
+
+- hosts: db_mongo
   gather_facts: yes
```
quartje, line 76: Here the same comment: There is the default option of installing mysql, and the extra option to be high available, and install galera. The latter could be behind a feature toggle as well.

provision.yml
dj-wasabi: But the role will only be executed when the host is part of the `loadbalancer_ha` group. So if that is empty, then it won't be installed.

provision.yml
dj-wasabi: Default `mysql` will be installed because the `storage` group has an host entry. The `galera` role will only be executed when there is an matching host entry in the `dbcluster` group.

provision.yml
quartje: On the loadbalancer_ha group all three roles need to be executed (haproxy, bind and keepalived) . In this setup, haproxy would not be installed on the production loadbalancers :)

provision.yml
quartje: I see. So different concepts are used to deploy the right roles on the right hosts. Does it make sense to unify that ? I see these three methods now:

1) Feature toggles (like iptables_enable and selfsigned_certs)
2) Host groups (and using that to exclude roles from being run)
3) Tags (where I believe that the "core" tag is to deploy a minimal set right?)

dj-wasabi: Yes, the "core" tag is the minimal set that will be deployed to have the basic set of services running.

In case of unifying it, will be difficult to answer that. The current way is mostly done to be backwards compatible, but also I don't have an exact way of how others are using it for their (production) environments. With for example for "mysql", a toggle `use_production_setup` can be used and if set to true, it will deploy the galera and keepalived.

But if you want to have seperate instances for running profile and engineblock, it might make more sense to have seperate groups for that and if the hostname is part of the group "profile", then "profile" will be installed on that host or not. If you want to use toggles for that, then you need to create/update the `host_vars/{{ inventory_hostname }}` but adding that toggle, or updating the inventory file and specifying the toggle over there (In which there are now 5 places (`template/vm/group_vars`, `template/vm/host_vars`, `template/vm/inventory`, `template/vm/secrets` and `group_vars`) where a property can be configured).

Also when using group, you can also make the "discovery" much more dynamic instead of using a property in group_vars that contains ip's or hostnames. With a `set_fact` task in a role you can create a new property and combine the hosts with an `,` seperator and use that in various configurations. So if a new host is added to that group, it will be automatically be picked up and configured on all places where you are using that property

With tags you just limit your ansible run and is always handy. If 1 role is updated, you only run that role. No toggle or group can fix that.

provision.yml
dj-wasabi: Done!

provision.yml
dj-wasabi: Applied the change.

provision.yml
dj-wasabi: Is now to be deployed with a toggle.

PR #303, opened 2020-11-16
Minor change that helps speedup deployment

dj-wasabi:

* Added pipelining (Enabling pipelining reduces the number of SSH operations required to execute a module on the remote server, by executing many ansible modules without actual file transfer.)

An average `./provision vm` on my Mac went from 780 seconds to 670 seconds.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,0 +1,2 @@
+[ssh_connection]
+pipelining=True


quartje (commented):
What happens when you add a ansible.cfg in the deployment directory? Will it merge your settings with the settings in /etc/ansible/ansible.cfg?

dj-wasabi:
> What happens when you add a ansible.cfg in the deployment directory? Will it merge your settings with the settings in /etc/ansible/ansible.cfg?

It will merge settings together when running Ansible, where the `ansible.cfg` in the repository wins in case it contains duplicate configuration settings.

quartje:
Seems like an excellent quick win then

PR #304, opened 2020-11-16
Do not log passwords to screen

dj-wasabi:

* Minor change to not print the (passwords) information to your screen which (other) people can see;


quartje (approved):
I've been hesitant to remove these, as it would also surpress the rest of the output. However, better do so, so let's merge and roll :)

PR #305, opened 2020-11-16
Combine several tasks into single task and added retry for network related tasks

dj-wasabi:

* Combined similar tasks into a single task;
* Added retries for tasks that does something with network related services, to prevent single network issues (Ansible best practice);
* Only run some tasks when other task is not changed;

### DIFF
--- a/roles/php/tasks/main.yml
+++ b/roles/php/tasks/main.yml
@@ -3,20 +3,20 @@
   yum:
     name: remi-release
     state: absent
+  register: php_packages_remi_remove
+  until: php_packages_remi_remove is succeeded
 
-- name: Install the remi-safe yum repo file
+- name: Install the remi-safe yum repo files
   copy:
-    src: "remi-safe.repo"
-    dest: "/etc/yum.repos.d/remi-safe.repo"
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
     owner: root
     mode: 0644
-
-- name: Install the remi RPM GPG key
-  copy: 
-    src:  "RPM-GPG-KEY-remi"
-    dest: "/etc/pki/rpm-gpg/RPM-GPG-KEY-remi"
-    owner: root
-    mode: 0644
+  with_items:
+    - src: remi-safe.repo
+      dest: /etc/yum.repos.d/remi-safe.repo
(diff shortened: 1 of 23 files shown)


roles/php/tasks/main.yml
```diff
@@ -81,10 +83,10 @@
...
     - "restart php72-php-fpm"
 
-- name: Enable php72-php-fpm
-  service:
-    name: php72-php-fpm
-    enabled: yes
+# - name: Enable php72-php-fpm
+#   service:
+#     name: php72-php-fpm
+#     enabled: yes
```
quartje, line 90: Why is this task commented?

quartje (commented):
Nice work! That seems to simplify a lot. 
Could you explain why the 'untils' are added? Is that Ansible best practice behaviour?

roles/php/tasks/main.yml
dj-wasabi: I commented it (Would actually wanted to remove it) but the handler is already enabling and (re)starting the service. So having a 2nd task which only does the enabling is a bit double.

quartje: Right, removing it should be ok than!

dj-wasabi:
> Nice work! That seems to simplify a lot.
> Could you explain why the 'untils' are added? Is that Ansible best practice behaviour?

Yes, it is considered to be a best practice. Assume that network related tasks could experience a timeout, hickup or any other temporary failure. With adding the `until` option, the task will be executed 3 times (Unless you specify more) before it actually fails and thus Ansible will try for example 3 times to install an package from the internet before it fails.

roles/php/tasks/main.yml
dj-wasabi: Yes, removed it. Thanks! 👍

quartje:
> > Nice work! That seems to simplify a lot.
> > Could you explain why the 'untils' are added? Is that Ansible best practice behaviour?
> 
> Yes, it is considered to be a best practice. Assume that network related tasks could experience a timeout, hickup or any other temporary failure. With adding the `until` option, the task will be executed 3 times (Unless you specify more) before it actually fails and thus Ansible will try for example 3 times to install an package from the internet before it fails.

:+1:

PR #308, opened 2020-11-23
Using springboot role and make use of a config to enable/disable easily functionality

dj-wasabi:
This PR contains the following chanhe:

* An property for both "gui" (`springboot_gui_services`) as "server" (`springboot_server_services`) is introduced with some basic keys, like `name`, `version` and `enabled`. When `enabled` is set to `true`, then the service will be deployed, otherwise not;
* For the "server" some more information is needed like which `port` it should be running, an `role` if the name of the role is something else than `name + '-server'`. Heapsizes can be configured if they should not have the default of `512m`;
* Removed the tasks and template directory from the "ROOT" dir;
* Removed the "dependencies" for the roles and moved it to the new `springboot` role;
* Added new springboot role that works on using a list as input to install/configure the springboot related services;
  * Each task name contains the name of the service, which could help debugging deployment;
  * An property `springboot_service_to_deploy` that eiher has `"all"` which will deploy all services found in either `springboot_gui_services` or `springboot_server_services` which has `enabled: true`. Can be overriden with the name of the service you want to deploy, with a comma `,` as a seperator if you want to deploy multiple;
  * In the `defaults/main.yml` the variables are used as a default and can be overridden with keys in the springboot_gui_services` or `springboot_server_services` properties;


environments/template/group_vars/template.yml
```diff
@@ -81,6 +81,132 @@ tls_star_cert: star.{{ base_domain }}.pem
...
+    enabled: true
+    version: "{{ teams_gui_version }}"
+  - name: pdp
+    enabled: true
+    version: "{{ pdp_gui_version }}"
+  - name: attribute-aggregation
+    enabled: true
+    version: "{{ attribute_aggregation_gui_version }}"
+  - name: dashboard
+    enabled: "{{ dashboard_install }}"
```
quartje, line 99: I think that this is not necessary anymore.

quartje (commented):
I've tested it on a working installation, and it seems to keep all the existing stuff intact, with only some minor things. 
I have a feeling that we might be overdoing things a bit by adding the whole springboot_gui_services and springboot_server_services to the individual environments. 

Would it be an idea to move those two to the defaults/main.yml of the springboot role? That would also eliminate the need to update your environment with every new springboot app.
It would require that the elements in it should become a variable (eg spring_boot_manage_enabled or something)

roles/springboot/tasks/springboot.yml
```diff
@@ -0,0 +1,46 @@
...
+  set_fact:
+    _springapp_service_name: "{{ springboot.name }}"
+    _springapp_type: "{{ springboot.type if springboot.type is defined else 'gui' }}"
+
+- name: "{{ springboot.name }}-{{ _springapp_type }} | Set facts (2)"
+  set_fact:
+    _springapp_min_heapsize: "{{ springboot.min_heapsize if springboot.min_heapsize is defined else springboot_min_heapsize }}"
+    _springapp_max_heapsize: "{{ springboot.max_heapsize if springboot.max_heapsize is defined else springboot_max_heapsize }}"
+    _springapp_random_source: "{{ springboot.random_source if springboot.random_source is defined else springboot_random_source }}"
+    _springapp_jar: "{{ springboot.jar if springboot.jar is defined else _springapp_service_name + '.jar' }}"
```
quartje, line 13: Currently the symlink is called $appname-current.jar. In order to prevent all these old symlinks laying around, is it possible to keep them the same?

environments/template/group_vars/template.yml
dj-wasabi: You mean the `dashboard` entry?

dj-wasabi:
> I've tested it on a working installation, and it seems to keep all the existing stuff intact, with only some minor things.
> I have a feeling that we might be overdoing things a bit by adding the whole springboot_gui_services and springboot_server_services to the individual environments.
> 
> Would it be an idea to move those two to the defaults/main.yml of the springboot role? That would also eliminate the need to update your environment with every new springboot app.
> It would require that the elements in it should become a variable (eg spring_boot_manage_enabled or something)

Yes, indeed. I will move it to the `defaults/main.yml`. So maybe this would also be the moment to have a switch/toggle file containing the enabling/disabling of the services?

roles/springboot/tasks/springboot.yml
dj-wasabi: That one is still created as before:
```
lrwxrwxrwx. 1 root  root         26 Nov 27 08:11 current -> /opt/teams/teams-gui-8.3.8
lrwxrwxrwx. 1 root  root         33 Nov 27 08:12 teams.jar -> /opt/teams/teams-server-8.3.8.jar
```

dj-wasabi:
> I've tested it on a working installation, and it seems to keep all the existing stuff intact, with only some minor things.
> I have a feeling that we might be overdoing things a bit by adding the whole springboot_gui_services and springboot_server_services to the individual environments.
> 
> Would it be an idea to move those two to the defaults/main.yml of the springboot role? That would also eliminate the need to update your environment with every new springboot app.
> It would require that the elements in it should become a variable (eg spring_boot_manage_enabled or something)

I have created a property `springboot_services_state` in the `defaults/main.yml`:

```
springboot_services_state:
  manage: true
  teams: true
  ...
```
So this one can be easily copied to group_vars when needed and both the `springboot_gui_services` as the `springboot_server_services` make use of this state property.

environments/template/group_vars/template.yml
dj-wasabi: I deleted the dashboard entry.

quartje (approved):

oharsta:
@dj-wasabi  Looks good! Nice improvement.

oharsta (approved):

PR #310, opened 2020-11-27
Able to create a galera cluster

dj-wasabi:

* Created a Molecule test that contains the installation of an MariaDB with Galera cluster;
* Made the `galera` role idempotent;
* The `galera` role is now able to build a working cluster with 2 or more nodes;
* Added pytests|testinfra tests to verify the Galera cluster;

### DIFF
--- a/molecule/mysql/converge.yml
+++ b/molecule/mysql/converge.yml
@@ -4,11 +4,6 @@
 
   vars:
     inventory_dir: "{{ lookup('env', 'MOLECULE_SCENARIO_DIRECTORY') }}/../../environments/vm"
-    mariadb_cluster_name: molecule
-    mariadb_cluster_password: secret
-    mariadb_root_password: secret
-    mariadb_backup_password: secret
-    galera_bootstrap_node: openconext-centos7-mysql
 
   pre_tasks:
     - name: Read vars from secrets file
@@ -16,23 +11,17 @@
     - name: Read vars from group_vars file
       include_vars: "../../environments/vm/group_vars/vm.yml"
 
-    - name: "Get ip current host"
-      shell: grep $(hostname) /etc/hosts | awk '{print $1}' | sort | tail -n 1
-      register: current_host_ip
-      changed_when: False
-
     - name: "Set fact for backend_ipv4"
       set_fact:
-        backend_ipv4: "{{ current_host_ip.stdout }}"
+        backend_ipv4: "{{ ansible_all_ipv4_addresses[0] }}"
+        mariadb_cluster_name: molecule
+        mariadb_cluster_password: secret
+        mariadb_root_password: secret
+        mariadb_backup_password: secret
+        galera_bootstrap_node: openconext-centos7-mysql
         galera_server_key: "{{lookup('file', inventory_dir + '/files/certs/galera/galera_server.key') }}"
         galera_client_key: "{{lookup('file', inventory_dir + '/files/certs/galera/galera_server.key') }}"
         galera_client_crt_name: "galera_server.pem"
-
-    - name: "Set specific fact for openconext-centos7-mysql-3"
-      set_fact:
-        arbiter_node: True
-      when:
-        - inventory_hostname == "openconext-centos7-mysql-3"
 
   roles:
     - role: mysql
--- a/molecule/mysql/converge.yml
+++ b/molecule/mysql/converge.yml
@@ -1,0 +1,52 @@
+---
+- name: Converge
+  hosts: all
+
+  vars:
+    inventory_dir: "{{ lookup('env', 'MOLECULE_SCENARIO_DIRECTORY') }}/../../environments/vm"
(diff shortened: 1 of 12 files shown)


molecule/mysql/converge.yml
```diff
@@ -0,0 +1,52 @@
...
+  pre_tasks:
+    - name: Read vars from secrets file
+      include_vars: "../../environments/vm/secrets/vm.yml"
+    - name: Read vars from group_vars file
+      include_vars: "../../environments/vm/group_vars/vm.yml"
+
+    - name: "Get ip current host"
+      shell: grep $(hostname) /etc/hosts | awk '{print $1}' | sort | tail -n 1
+      register: current_host_ip
+      changed_when: False
```
quartje, line 22: Is it not possible to get the host IP address using gather_facts with molecule?

molecule/mysql/prepare.yml
```diff
@@ -0,0 +1,121 @@
...
+        src: "{{ item.dest }}"
+        dest: "{{ item.src }}"
+      with_items: "{{ copy_tls_ca }}"
+
+    - name: create host CSR signing key
+      openssl_privatekey:
+        path: /root/example_com_host_key.pem
+      register: example_com_key
+
+    - name: create the CSR for the LDAP server
```
quartje, line 74: LDAP should be Galera?

roles/galera/defaults/main.yml
```diff
@@ -7,6 +7,8 @@ galera_bootstrap_node: ""
 mariadb_cluster_user: galera_sst_user
 mariadb_backup_user: backup_user
 
+percona_url_rpm: https://mirrors.cloud.tencent.com/percona/release/7/RPMS/noarch/percona-release-1.0-25.noarch.rpm
```
quartje, line 10: Why not use the official Percona YUM repo?

roles/galera/tasks/cluster_nodes.yml
```diff
@@ -95,14 +96,21 @@
...
+  register: mariadb_new_installation
+
 - name: Create state transfer key
-  shell: cat /etc/pki/mysql/{{galera_server_key_name }} /etc/pki/mysql/{{galera_server_crt_name }} > /etc/pki/mysql/{{ galera_sst_crt_name }}
+  shell: awk 'FNR==1{print ""}1' /etc/pki/mysql/{{ galera_server_key_name }} /etc/pki/mysql/{{galera_server_crt_name }} > /etc/pki/mysql/{{ galera_sst_crt_name }}
+  args:
+    creates: "/etc/pki/mysql/{{ galera_sst_crt_name }}"
   when:
     - galera_ssl | bool
-  changed_when: false
```
quartje, line 105: Is the awk command idempotent?

roles/galera/defaults/main.yml
dj-wasabi: The url that is part of the role didn't work, it gave a 404.

roles/galera/templates/server.cnf.j2
```diff
@@ -95,6 +94,5 @@ wsrep_log_conflicts=ON
 [mariadb-10.0]
 
 [sst]
-encrypt = 2
```
quartje, line 98: Not needed anymore?

roles/galera/tasks/cluster_nodes.yml
dj-wasabi: The `awk` command itself not, but task is never executed because the task is checking to see if file `/etc/pki/mysql/{{ galera_sst_crt_name }}` exist. If that is the case, the task is skipped.

roles/galera/templates/server.cnf.j2
dj-wasabi: No, it complains with an warning that the property is deprecated.

molecule/mysql/prepare.yml
dj-wasabi: Ah yes, was a copy/paste. Will update it with next commit. 👍

molecule/mysql/converge.yml
dj-wasabi: I had issues before in using the gather_facts for getting the last ip (basically the network that Molecule is creating), so since then I use these tasks with every Molecule test that I have created.

roles/galera/defaults/main.yml
quartje: Or maybe use this one? 

https://repo.percona.com/yum/percona-release-latest.noarch.rpm

Mirror RPM URLs tend to disappear ...

roles/galera/tasks/cluster_nodes.yml
quartje: :+1:

roles/galera/defaults/main.yml
dj-wasabi: Yes, I am testing this locally now. Will commit and push with a minute. 👍

PR #311, opened 2020-11-30
Build Docker image

dj-wasabi:
* Using the `-t core` to only deploy the "core" related services with Travis;
* Create an Docker image from the Travis docker run and tag it;
* Commented, the `docker push` action;
* Added a `docker-compose` file so it can make use of the Docker image;

Before this can be merged (If we want to), the following should be provided/clear:
* Once the Ansible run is done in the Docker image, an Docker image with the size of 2,5GB is created. Is this acceptable or not?
* If acceptable what would be the correct name of the Docker image;
* Does it needs to be pushed to the offical Docker Hub? If so, be aware of the rate limits for the Docker hub (https://www.docker.com/increase-rate-limits);
* When the correct name of the Docker image is clear, this PR should be updated so it will be pushed;
* Once the PR is updated, probably some changes in Travis needs to be applied so it can do a `docker login` so the Docker image can be pushed to the chosen registry;

Small note:
Creating an Docker image has an size of 2,5GB. Creating 2 Vagrant boxes, 1 for LB has 600MB for an size and 1 for APP which has an size of 1.1GB. So, the Vagrant boxes together are smaller than the single Docker image. So, is using a single Docker still the way forward, or should we do something similar for Vagrant?


quartje:
Seems to be working, but two issues need to be resolved:

* in the travis-build.sh we enable oidncng (see the three lines below  # Enable oidc-ng ). When doing a -t core those tasks should not be performed
* The Haproxy role is quite complicated, and in the travis setup, it used to build a Haproxy config with two frontends: One listening on 127.0.0.1 and one on 127.0.0.2. The latter is meant for applications that should be firewalled of the internet (like eb-api and manage). Now the first ip is changed to 0.0.0.0, and the second stays at 127.0.0.2. The thing is that it makes the applications runnin g behind 127.0.0.2 unreachable from outside the Docker image (since the 443 - 443 port forwarding only makes the 0.0.0.0 apps reachable). I'd like to suggest that have all apps run behind 0.0.0.0 and only eb-api behind 127.0.0.2 (simply by removing "restricted: yes" from the respective haproxy_applications in the groups_vars except for engine-api).

dj-wasabi:
> * When doing a -t core those tasks should not be performed

I have applied the suggested changes.

PR #312, opened 2020-12-02
Feature/java no more java7

quartje:
We still install Java 7, which is not used anymore. Removing them, it also saves a lot of space when using Docker


quartje:
Wrong branch to merge into. Closing and reopening on the correct branch

PR #314, opened 2020-12-02
Java: Remove the java 7 install, it's not used anymore

quartje:
We still install Java 7, which is not used anymore. Removing them, it also saves a lot of space when using Docker


oharsta (approved):
I think it is time we add the latest JDK here and slowly start migrating some apps.

PR #315, opened 2020-12-02
Feature/provision apps from role to manage

quartje:
This will provision the SAML SP and IdP entities to Manage, directly from their respective roles.
It was necessary to rewrite the order of the provision.yml playbook, since you'd first need manage, then eb and oidcng (since manage needs those to push the entities to) and then the other apps.
I've configured the bare minimum to get this working, more could be added later on (like logo's)

This change will ensure that the provisioning can take a more modular approach, since only the needed apps are provisoned.
It also eliminates the role vm_only_provision_manage_eb, which was a Mongo dump that was restored. It was not maintained, and cumbersome to add apps to it. That is now much easier.


oharsta:
@quartje Looks good!

oharsta (approved):

PR #316, opened 2020-12-18
Remove oidc, authz* installation

thijskh:
First start at authz*/oidc cleanup.


thijskh:
Do we want to remove `roles/{authz*,oidc}` also at this point?

quartje:
Nice cleanup! I think it's safe to remove the roles directories as well. It gives a clear signal that there is no more support for these apps.

quartje (approved):

quartje (approved):
I've removed the secrets for oidc as well (and cleaned up some leftover oidcng stuff). 
The monitoring tests config file also still contained oidc leftovers, I've cleaned that up as well. Seems good to go!

PR #318, opened 2020-12-27
Ability to group services with an alias

oharsta:
With the introduction of springboot services the 'tag' functionality was lost. This reinstates this  functionality with an optional alias attribute to group services.


quartje:
Nice addition!

PR #319, opened 2021-01-04
Compatibility and performance improvements

MKodde:
Work in progress.

Todo: if optimizations are found to be actual improvements, also apply similar directives for profile/voot/EB-api

### DIFF
--- a/roles/engineblock/templates/engine.conf.j2
+++ b/roles/engineblock/templates/engine.conf.j2
@@ -48,4 +48,9 @@ Listen {{ apache_app_listen_address.engine }}:{{ loadbalancing.engine.port }}
     Include ssl_backend.conf
     {% endif %}
 
+    ExpiresActive on
+    ExpiresByType font/* "access plus 1 year"
+    ExpiresByType image/* "access plus 6 months"
+    ExpiresByType text/css "access plus 1 year"
+    ExpiresByType text/js "access plus 1 year"
 </VirtualHost>
--- a/roles/httpd/templates/httpd.conf_httpd24.j2
+++ b/roles/httpd/templates/httpd.conf_httpd24.j2
@@ -265,6 +265,8 @@ LogLevel warn
     #
     AddType application/x-compress .Z
     AddType application/x-gzip .gz .tgz
+    AddType font/woff2 .woff2
+    AddType image/x-icon .ico
 
     #
     # AddHandler allows you to map certain file extensions to "handlers":
@@ -297,6 +299,8 @@ LogLevel warn
 # directive:
 #
 AddDefaultCharset UTF-8
+AddCharset UTF-8 .css
+AddCharset UTF-8 .js
 
 <IfModule mime_magic_module>
     #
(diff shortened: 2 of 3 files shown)


MKodde:
@thijskh and @quartje 

This PR mainly satisfies findings from the `webhints` extension. They do follow common best practices so I think some of these changes make sense. 

My questions to you are: 
 - Is this the way to go? Increasing cache expiration times on commonly used file types, and adding the cache busting `?v=assetversoin` on the EB Twig template side.
 - Do we want to set Content-Type associations 'globally' in the Apache config? Or rather set them application specifically?

quartje:
I think that these changes could help speeding up things a bit. However, we need to be sure that the file types that get cached are never updated. I think that it works fine for the files that get the ?assetversion parameter, but for things like images I'm not so sure that that is the case. 
I don't think there's any harm in setting the Content-Type associations globally

MKodde:
Agreed, adding a year long cache on all `png` images is somewhat harsh, on the other hand we are able to cache-bust them using the `v` query parameter, combined with the `assetVersion`. I agree with you that is not the nicest way to go about dealing with images. 

We could limit this change to non-image resources to get a good poldermodel?

@thijskh any thoughts from your side?

roles/engineblock/templates/engine.conf.j2
```diff
@@ -48,4 +48,14 @@ Listen {{ apache_app_listen_address.engine }}:{{ loadbalancing.engine.port }}
     Include ssl_backend.conf
     {% endif %}
 
+    <IfModule mod_expires.c>
```
thijskh, line 51: This means that if you don't enable the module it will silently just not work as intended. Would it be more predictable to remove the guard so Apache will alert you if you forgot to enable the module?

roles/engineblock/templates/engine.conf.j2
```diff
@@ -48,4 +48,14 @@ Listen {{ apache_app_listen_address.engine }}:{{ loadbalancing.engine.port }}
...
+      ExpiresActive on
+      ExpiresByType image/x-icon "access plus 1 year"
+      ExpiresByType font/woff2 "access plus 1 year"
+      ExpiresByType text/css "access plus 1 year"
+      ExpiresByType text/js "access plus 1 year"
+      ExpiresByType image/png "access plus 6 months"
```
thijskh, line 57: Or `image/*` `font/*`? Or is that too bold?

thijskh:
I don't see the problem with images. They are also referenced in twig with an asset version (or should be if they are not).

For the idp logos we currently use static.example.org. We should add similar headers there for optimal result. And ensure in our process that updating an idp logo on static means changing its name.

roles/engineblock/templates/engine.conf.j2
MKodde: Bold is better! Going to test if this introduces additional issues on EB. I don't think so, otherwise we would have noticed with the inspections..

roles/engineblock/templates/engine.conf.j2
MKodde: Good point, should we advertise the dependency on this module as a requirement somewhere in docs?

MKodde:
Thanks @thijskh your feedback was applied to the change-set.

I also applied the change to Profile. I was not brave enough to apply it to other components.

thijskh (approved):

thijskh:
If we want to be brave, we move this to a snippet in the httpd role instead so we can include it for all apps.

quartje (approved):

PR #320, opened 2021-01-06
Set httponly flag on lang cookie

MKodde:
The cookie is used exclusively by the PHP application, so a http_only flag set to true is in order here. This change is also applied in parameters distfile in EB

See: https://www.pivotaltracker.com/story/show/176287495
See: https://github.com/OpenConext/OpenConext-engineblock/pull/1056

### DIFF
--- a/roles/engineblock/templates/parameters.yml.j2
+++ b/roles/engineblock/templates/parameters.yml.j2
@@ -150,7 +150,7 @@ parameters:
     cookie.secure: true
     cookie.locale.domain: .{{ base_domain }}
     cookie.locale.expiry: 5184000
-    cookie.locale.http_only: false
+    cookie.locale.http_only: true
     cookie.locale.secure: true
 
     ## UI settings


thijskh:
"problem" with the lang cookie is that it is platform global (i.e. `.surfconext.nl` as its domain). So we must be sure that other platform components that use the cookie will also not use it in JavaScript.

MKodde:
Good point! I can speak for myself that the other PHP applications we maintain do not use the lang cookie in JavaScript. But I'm not sure about the Java (dark) side of things ;) 

Maybe @oharsta knows this from the top of his head?

tvdijen:
PDP & Teams rely on it for sure

tvdijen:
And if I may add to the pivotal-story, there's also a rememberChoice cookie that's being set secure=false & httponly=false

MKodde:
@tvdijen do those applications use these cookies in Java or JavaScripts? If it is the latter, that would be reason not to make them http_only.

Regarding the rememberChoice cookie. See a [recent change](
https://github.com/OpenConext/OpenConext-engineblock/pull/1056/files ) in EB, targetted for version `6.4.1` later today.

tvdijen:
Javascript @MKodde

MKodde:
In that case, we should not make the lang cookie Http Only! Thanks for your review @tvdijen

PR #321, opened 2021-01-11
Configure defaultIdp in parameters.yml

MKodde:
@thijskh or @quartje what's a neat way to get the eduId entityId/clientId into the parameters.yml template? I looked at the defaults configured in the group_vars, but to no avail.

Who can throw me a lifeline?


thijskh:
How about e5eaa628646394c1f28fcc923d42a7873c04e191?

PR #323, opened 2021-01-17
Feature/split RP RS

oharsta:


roles/manage-server/files/metadata_configuration/oauth20_rs.schema.json.j2
```diff
@@ -0,0 +1,181 @@
...
+        },
+        "^OrganizationDisplayName:({{ supported_language_codes | replace(',','|') }})$": {
+          "type": "string",
+          "info": "The friendly name of the organization. e.g. University of Harderwijk."
+        },
+        "^NameIDFormats:([0-2]{1})$": {
+          "$ref": "#/definitions/NameIDFormat",
+          "multiplicity": 3,
+          "info": "The NameIDFormat(s) this service provider supports. This can be transient, persistent or unspecified."
+        },
```
quartje, line 126: Do we need the NameIDFormats in addition to the NameIDFormat?

roles/manage-server/files/metadata_configuration/oauth20_rs.schema.json.j2
```diff
@@ -0,0 +1,181 @@
...
+          ],
+          "multiplicity": 4,
+          "sibblingIndependent": true,
+          "info": "From the pull down, select the type of contact."
+        }
+      },
+      "required": [
+        "name:en",
+        "secret",
+        "scopes"
```
quartje, line 169: Is this really required?

oharsta: Good question. We need 1 name and without secret the RS can not introspect. Scopes we can do without I guess.

roles/manage-server/files/metadata_configuration/oauth20_rs.schema.json.j2
oharsta: No, we don't. NameIDFormat is sufficient.

PR #330, opened 2021-03-09
Feature/php use opcache

quartje:
opcache needs to be explicitly installed in order to be available.
Initial tests show that this will make Engineblock and Profile an order of magnitude faster


thijskh (approved):
Yes, this is useful, great find. I was surprised that it was not enabled because I was under the impression that this was the case since PHP 5.5+. It seems it's bundled since that version but it depends on your OS whether it's actually enabled. Given that it's enabled by default on other mainstream distributions I also see not many risks.

PR #333, opened 2021-03-12
Feature/haproxy add maps

quartje:
Some logic has been moved from the configuration file into lists and maps.
* Allowed vhosts are in a seperate file
* The map host header / backend has been placed in a mapfile
* Redirects are placed in a mapfile

This simplifies the configuration, and also makes it possible to change the maps and acls on the fly.

### DIFF
--- a/roles/haproxy/tasks/main.yml
+++ b/roles/haproxy/tasks/main.yml
@@ -61,6 +61,17 @@
     group: lbops
     mode: 0775
 
+- name: Create haproxy config map and acl directories
+  file:
+    path: "/etc/haproxy/{{ item }}"
+    state: directory
+    owner: root
+    group: lbops
+    mode: 0755
+  with_items:
+    - acls
+    - maps
+
 - name: Enable haproxy
   service:
     name: haproxy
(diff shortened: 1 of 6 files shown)


roles/haproxy/tasks/main.yml
```diff
@@ -144,15 +155,36 @@
...
 - name: Place file with list of browsers that do not accept samesite=none cookies
   copy:
     src: nosamesitebrowsers.lst
-    dest: /etc/haproxy/nosamesitebrowsers.lst
-    owner: root
-    group: root
-    mode: 0644
-  notify: 
-    - "reload haproxy" 
+    dest: /etc/haproxy/maps/nosamesitebrowsers.lst
```
thijskh, line 187: Why is this not part of the previous task? Only because it’s not a template? Because the permissions and reload behaviour now differ.

thijskh (approved):

roles/haproxy/tasks/main.yml
quartje: Good catch, I've combined the two into one task.

thijskh: Moet je de file wel renamen naar nosamesitebrowsers.lst.j2

quartje: Thanks for reminding me :)

I've made it into two tasks again, the build is now green, and the permissions are what they should be. 
Merging ..

PR #335, opened 2021-03-15
Haproxy: Add hitless reload options

quartje:
This will allow a reload of Haproxy without losing any connections,
under heavy loads. See: https://www.haproxy.com/blog/truly-seamless-reloads-with-haproxy-no-more-hacks/

This might be designed for very high loads, but it doesn't do any harm anyway.


thijskh:
Don’t you also need to change the systemd invocation to make use of this?

quartje:
> Don’t you also need to change the systemd invocation to make use of this?

The default systemd startup options already include the -W option, which is indeed needed. Our haproxy build also supports this default behaviour

thijskh (approved):

PR #340, opened 2021-05-04
iptables: Add specific drop rule at the end of the other rules

quartje:
There are no drop rules present in the iptables and ip6tables rules.


thijskh (approved):
As we discussed the defaults at the start of the file already DROP unwanted traffic. However, I think these rules are an improvement as they explicitly reject traffic which is both fail-fast and explicit behaviour.

PR #342, opened 2021-06-07
Make supported acr_values in OIDC playground match those configured for OIDCNG

thijskh:
Instead of using the hardcoded list that ships with oidc-playground jar


quartje (approved):
Ah the old "reverting the reverted commit commit" :)

oidc-playground read the well-known openid-configuration before, but these two can contain different values now, since oidcng transparently passes the acr value (and we can in theory support all the ACR values that exist)

PR #343, opened 2021-06-29
Feature/haproxy block http methods

quartje:
This adds the ability in Haproxy to whitelist only allowed http methods. 

The current list is:
GET HEAD OPTIONS POST PUT DELETE

Other methods are not needed by the OpenConext stack.

### DIFF
--- a/roles/haproxy/defaults/main.yml
+++ b/roles/haproxy/defaults/main.yml
@@ -1,4 +1,6 @@
 tls_backend_ca: /etc/pki/haproxy/backend.{{ base_domain }}.pem
 haproxy_cookie_max_idle: 3600
 haproxy_metricbeat: False
+# The default http methods whitelist is meant to disable methods like TRACE or DEBUG.
+# OPTIONS is mainly used for CORS requests in OIDC-NG (like SPA apps)
 haproxy_supported_http_methods: "GET HEAD OPTIONS POST PUT DELETE"
--- a/roles/haproxy/defaults/main.yml
+++ b/roles/haproxy/defaults/main.yml
@@ -1,4 +1,3 @@
 tls_backend_ca: /etc/pki/haproxy/backend.{{ base_domain }}.pem
-haproxy_removal_applications: []
 haproxy_cookie_max_idle: 3600
 haproxy_metricbeat: False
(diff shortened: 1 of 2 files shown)


roles/haproxy/templates/haproxy_frontend.cfg.j2
```diff
@@ -83,6 +86,9 @@ frontend internet_restricted_ip
     capture request header User-agent len 256 
     # Put the useragent header in a variable, shared between request and response. 
     http-request set-var(txn.useragent) req.fhdr(User-Agent)
+    # The ACL below makes sure only supported http methods are allowed
+    acl valid_method method {{ haproxy_supported_http_methods }}
+    http-request deny if !valid_method
```
thijskh, line 91: Returning a 405 seems to match the situation better than 403.

roles/haproxy/defaults/main.yml
```diff
@@ -1,4 +1,4 @@
 tls_backend_ca: /etc/pki/haproxy/backend.{{ base_domain }}.pem
-haproxy_removal_applications: []
 haproxy_cookie_max_idle: 3600
 haproxy_metricbeat: False
+haproxy_supported_http_methods: "GET HEAD OPTIONS POST PUT DELETE"
```
thijskh, line 4: Should we somehow document how we arrived at this list? Who uses what? How do we know that perhaps we can remove one method at some point?

roles/haproxy/templates/haproxy_frontend.cfg.j2
quartje: You're right. Fixed it!

roles/haproxy/defaults/main.yml
quartje: I'm not very sure where we should document this, and how that list should be kept up to date. How would you propose that?
Pragmatically speaking, I'd say that by whitelisting these most common methods we at least keep out any other methods like TRACE. If needed, apps can themselves keep a more restricted list. It doesn't seem very realistic that any of these methods would disapear in the near future anyway.

thijskh: Someone in the (near) future will come and ask "why these specific methods" or "why is `OPTIONS` in there". Maybe it is already sufficient to just add a comment that says that this is the full list minus things we know we never need.

quartje: I've added a comment in the defaults.yml.

thijskh (approved):

PR #344, opened 2021-07-01
Feature/haproxy block excessive requests

quartje:
A few changes are present in this PR:

I've rewritten the collection of request rates so that each sticky table is in its' own backend: There is only one such definition allowed per front- or backend. I've also created a seperate config file for the sticky tables. Furthermore, ipv6 addresses are now taken into account as well. 

Up until now, the request rate per ip address was only logged, no blocklist acl was yet in place. I've added it here, and the default is 1000 requests per 10s (total connections to all applications behind the loadbalancer). Some NATted ip addresses are capable of generating over 500 valid requests per 10s. Logging showed that all requests over 1000 per 10s are bots. When the amount of requests drops below 1000 per 10s requests will be allowed again. The rate limit is configurable by setting `haproxy_max_request_rate`

Of course, the measure above is course grained, but meant to keep out abusive bots generating too many requests per second.
I've added a second sticky table to log the amount of requests per ip address/URL path combination: the path is everything after the hostname, and before the first "?" in a URL. The idea is to block abusers that generate lots of requests to a single path. For now, only logging is added to get an idea of what is considered normal, and what is typical abusive behaviour so we can block it with a 429 in the future.

### DIFF
--- a/roles/haproxy/README.md
+++ b/roles/haproxy/README.md
@@ -85,6 +85,25 @@ certs: A list of certificates that you use for https. That list contains these p
  key_content: You can add your https key in this parameter, and add it to your secrets. See environments/vm/secrets/vm.yml for an example
  crt_name: The certificate is placed in your environments directory: In files/certs/. The filename should match "crt_name"
 
+## Rate limiting
+The frontend has a limit that is configurable for the amount of requests per 10s. The rationale is that you want to stop sudden bursts of requests fired by clients, and you want to be quick enough to block it, before your backends are saturated. The default is 1000 requests per 10s. In our logs, we observed some NAT gateways of larger instutitions that were able to generate more than 500 legitimate requests per 10s. All rates above 1000 per 10s seemed to be coming from bots, scrapers, or misbehaving clients.
+If you want to exclude ip's (either in cidr notation, or single ip addresses) from the limit, you can add them to your configuration under ```haproxy_allowlistips```. Suppose you want to put the ip address 1.1.1.1 and the ip addres 2.2.2.2 to the allowlist, you can do so by putting the following in your environment:
+```
+haproxy_allowlistips:
+  - 1.1.1.1
+  - 2.2.2.2
+```
+
+You can also do this on a running loadbalancer by using entering this command:
+```
+echo "add acl /etc/haproxy/acls/allowedips 1.1.1.1" | socat unix-connect:/var/lib/haproxy/haproxy.stats stdio
+echo "add acl /etc/haproxy/acls/allowedips 2.2.2.2" | socat unix-connect:/var/lib/haproxy/haproxy.stats stdio
+```
+Please note that such a change does not survive a restart of haproxy. Adding them to the acl and deploying haproxy makes it permanent. 
+
+
+
+
 
 
 
--- a/roles/haproxy/templates/haproxy_frontend.cfg.j2
+++ b/roles/haproxy/templates/haproxy_frontend.cfg.j2
@@ -52,17 +52,21 @@ frontend local_ip
     option httplog
     capture request header User-agent len 256
     capture request header X-TLS-Client len 256
+    # Allowlist ACL
+    acl allowlist src -f /etc/haproxy/acls/allowedips.acl
     # Measure and log the request rate per 10s
     http-request track-sc0 src table st_httpreqs_per_ip
     http-request capture sc_http_req_rate(0) len 4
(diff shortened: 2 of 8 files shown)


roles/haproxy/README.md
```diff
@@ -85,6 +85,25 @@ certs: A list of certificates that you use for https. That list contains these p
  key_content: You can add your https key in this parameter, and add it to your secrets. See environments/vm/secrets/vm.yml for an example
  crt_name: The certificate is placed in your environments directory: In files/certs/. The filename should match "crt_name"
 
+## Rate limiting
+The frontend has a limit that is configurable for the amount of requests per 10s. The rationale is that you want to stop sudden bursts of requests fired by clients, and you want to be quick enough to block it, before your backends are saturated. The default is 1000 requests per 10s. In our logs, we observed some NAT gateways of larger instutitions that were able to generate more than 500 legitimate requests per 10s. All rates above 1000 per 10s seemed to be coming from bots, scrapers, or misbehaving clients.
+If you want to exclude ip's from the limit, you can add them to your configuration under ```haproxy_allowlistips```. Suppose you want to put the ip address 1.1.1.1 and the ip addres 2.2.2.2 to the allowlist, you can do so by putting the following in your environment:
```
thijskh, line 90: IPs? Or cidrs?

roles/haproxy/templates/haproxy_frontend.cfg.j2
```diff
@@ -49,10 +52,17 @@ frontend local_ip
...
+    http-request track-sc0 src table st_httpreqs_per_ip
     http-request capture sc_http_req_rate(0) len 4
+    # Create an ACL when the request rate exceeds {{ haproxy_max_request_rate }} per 10s and deny them
+    acl exceeds_max_request_rate_per_ip sc_http_req_rate(0) gt {{ haproxy_max_request_rate }}
+    http-request deny deny_status 429 if exceeds_max_request_rate_per_ip !allowlist
+    # Measure and log the request rate per path and ip 
+    http-request track-sc1 base32+src table st_httpreqs_per_ip_and_path
+    http-request capture sc_http_req_rate(1) len 4
     # Deny the request when no valid host header is used
     http-request deny if ! valid_vhost
```
thijskh, line 67: Is there a rationale for the ordering of the checks? I would do them from simplest to less simple: check if method is even allowed, check if hostname even exists (basic sanity - we can reject those out of hand anyway. Only then check if there’s too many of them. It also makes the generated error more specific.

roles/haproxy/templates/haproxy_frontend.cfg.j2
```diff
@@ -106,10 +119,17 @@ frontend localhost_restricted
...
-    stick-table type ip size 1m expire 10s store http_req_rate(10s)
-    http-request track-sc0 src
+    # Allowlist ACL
+    acl allowlist src -f /etc/haproxy/acls/allowedips.acl
+    # Measure and log the request rate per 10s
+    http-request track-sc0 src table st_httpreqs_per_ip
     http-request capture sc_http_req_rate(0) len 4
+    # Create an ACL when the request rate exceeds {{ haproxy_max_request_rate }} per 10s and deny them
+    acl exceeds_max_request_rate_per_ip sc_http_req_rate(0) gt {{ haproxy_max_request_rate }}
+    http-request deny deny_status 429 if exceeds_max_request_rate_per_ip !allowlist
```
thijskh, line 129: How useful is it to apply these things to vhosts only accessible from trusted networks?

roles/haproxy/README.md
quartje: both, I've updated the documentation to reflect that

roles/haproxy/templates/haproxy_frontend.cfg.j2
quartje: If they're trusted, and well behaved, there is no issue. If they do generate requests in these numbers, I'd say they should be blocked as well

roles/haproxy/templates/haproxy_frontend.cfg.j2
quartje: No second thought on the ordering was given. It makes more sense indeed to do it in the proposed order. I've changed it

roles/haproxy/README.md
thijskh: I don’t see that change yet?

quartje: Somehow it got in my Git stash. I've committed it now!

thijskh (approved):
Good work, maybe remove the debug statement

PR #345, opened 2021-07-06
Feature/haproxy server maxconn variable

quartje:
This change makes the amount of concurrent requests to a backend configurable. After reaching this limit, Haproxy will queue the requests. It makes sure that the backends are not saturated 
The default is 35 concurrent connections (as is the default currently in the php fpm servers as well). It can be overriden on a per application basis in haproxy_applications, setting the maxconn variable per app. If it's not defined, the default of 35 is used.

### DIFF
--- a/roles/haproxy/templates/haproxy_backend.cfg.j2
+++ b/roles/haproxy/templates/haproxy_backend.cfg.j2
@@ -34,7 +34,7 @@
   cookie HTTPSERVERID insert nocache indirect httponly secure maxidle {{ haproxy_cookie_max_idle }}
 
   {% for server in application.servers %}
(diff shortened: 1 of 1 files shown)


thijskh:
How would we know that the limit is reached?

quartje:
> How would we know that the limit is reached?

It can be checked in the logs (or ELK if you have that installed). You check for srv_queue. From the documentation:
```
- "srv_queue" is the total number of requests which were processed before
    this one in the server queue. It is zero when the request has not gone
    through the server queue. It makes it possible to estimate the approximate
    server's response time by dividing the time spent in queue by the number of
    requests in the queue. It is worth noting that if a session experiences a
    redispatch and passes through two server queues, their positions will be
    cumulative. A request should not pass through both the server queue and the
    backend queue unless a redispatch occurs.
```

thijskh (approved):

PR #347, opened 2021-07-14
Haproxy: Disable tls1.0 and 1.1 in the public frontend

quartje:
TLS1.0 and 1.1 were already disabled in the restricted backend. This change allows disabling it in the public frontend as well. The default is still to have them enabled. Setting `haproxy_disable_tls10_and_11` to `True` in your variables will disable them


thijskh (approved):
I was about to say that I prefer "non-negative" options, but given that the actual haproxy options are also negatively formulated it seems fitting in this way.

Maybe add the default = true to the templates for new installs as a public service?

quartje:
I've changed the default to "True".

PR #348, opened 2021-09-14
specify sysadmin user to Check, Upload and Push

domgon:
We add some issues on this role when installing Voot. In our case Ansible was trying to Upload the entity to Manage using the user **sysadmin** but the user was **stats** that we add in the past for the statistics user to get access to Manage. On that role the order of **manage.apiUsers** variable matters. On our case the variables order on **manage.apiUsers** was different. The third item **[2]** on the array was not **sysadmin**, but **stats**. To avoid to miss the correct user to do that operation on that role, we suggest to specify the user and get directly his password


quartje:
Thanks Domingos, this simplifies things. sysadmin user is present in the template, so I guess it should be available in all installations.

PR #350, opened 2021-09-14
Add the role haproxy_acls

quartje:
This role allows you to edit the allow and blocklists of haproxy in the live running instances, using the Haproxy stats socket.


roles/haproxy_acls/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Create a blocklist and check first if it's not already on the list
+  shell: echo "show acl /etc/haproxy/acls/blockedips.acl" | socat unix-connect:/var/lib/haproxy/haproxy.stats stdio |awk '{print $2}' |sort |uniq | grep {{ item }}
```
thijskh, line 3: Why sort and uniq if you will be grepping anyway?

roles/haproxy_acls/tasks/main.yml.eenhost
```diff
@@ -0,0 +1,8 @@
+- name: set weights for haproxy hosts
```
thijskh, line 1: Not an intended part of this PR?

roles/haproxy_acls/tasks/main.yml.sa
```diff
@@ -0,0 +1,52 @@
+---
```
thijskh, line 1: Not an intended part of this PR?

roles/haproxy_acls/tasks/main.yml
thijskh: Why is the create+check step necessary? Can't you just always add it?

quartje:
I've simplified the whole thing. No more live stats socket adaptations to running haproxy, but a simple ACL file change.
Additionally, I've given the lbops user the sudo permission to reload Haproxy. Since these are seamless, the whole operation can be safely performed, even during peak hours

thijskh (approved):

PR #353, opened 2021-09-17
Fix Java heap memory settings

quartje:
When the java roles were moved to the springboot role, the java memory settings were not correctly put into the correct variables. This fixes this issue

### DIFF
--- a/roles/springboot/tasks/main.yml
+++ b/roles/springboot/tasks/main.yml
@@ -36,7 +36,7 @@
 - name: "Install Springboot Server services"
   include_tasks: "springboot.yml"
   with_items: "{{ springboot_server_services }}"
-  no_log: true
+    #  no_log: true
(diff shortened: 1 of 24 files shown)


roles/springboot/tasks/main.yml
```diff
@@ -36,7 +36,7 @@
 - name: "Install Springboot Server services"
   include_tasks: "springboot.yml"
   with_items: "{{ springboot_server_services }}"
-  no_log: true
+    #  no_log: true
```
thijskh, line 39: Why is this?

thijskh (approved):

roles/springboot/tasks/main.yml
quartje: Good catch, that was for my local debugging purposes, I've reverted it now.

PR #354, opened 2021-09-17
Shibboleth: Make sure only one server run cleanup jobs

quartje:
In a multiserver setup with a shared database, you could have multiple shib instances trying to run the same delete operation while running cleaning up tasks. You can now set shibboleth_db_cleanup_interval to 0 on hosts where you don't want to run the db cleanup tasks, leaving one master to run these tasks.

### DIFF
--- a/roles/shibboleth/defaults/main.yml
+++ b/roles/shibboleth/defaults/main.yml
@@ -4,3 +4,4 @@ shibboleth_metadata_sources:
 shibd_sp_crt_not_in_inventory: false
 mariadb_odbc_version: 3.0.8
 shibboleth_database_backend: false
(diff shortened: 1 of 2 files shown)


thijskh (approved):
I have found out by studying the Shibboleth source code (as it's not in the docs) that indeed a value <= 0 disables cleanup. So 👍

quartje:
Heh, I found it in some mailthread when DuckDuckgo'ing  the configuration parameter :)

thijskh:
https://shibboleth.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=2065334638&selectedPageVersions=14&selectedPageVersions=15

PR #356, opened 2021-09-22
Add EngineBlock consent remove feature toggle

MKodde:
This feature is configured in the parameters.yml so it is included in
the parameters.yml jinja template.

https://www.pivotaltracker.com/story/show/179490276


quartje (approved):
Looks good to me, with sane defaults.

PR #359, opened 2021-10-08
Haproxy: Protect /internal path

quartje:
This change allows protecting all paths behind the loadbalancer that start with /internal/ for access only from an ip allowlist. This will be used to protect backend APIs like the deprovision API, or the API in Teams for the SPdashboard that have been secured with http basic authentication only. 
The default is to allow access from anywhere, additional documentation will be added to the wiki to make sure first time deployers understand how to secure their OpenConext installation.


thijskh:
And then we change monitor bundle to expose `/internal/health` and `/internal/info`?

PR #365, opened 2021-11-15
Teams spd api limit access

thijskh:


quartje:
Nice work! The Apache 2.4 style authorization, combined with some of the rewrites make it a lot easier to understand what is actually happening.

PR #368, opened 2021-11-24
Make use of globally defined Content-Security-Policy flavours.

thijskh:
Also define static_vhost globally. Environments may override this. So e.g. test2 can define it as "static.surfconext.nl" (not static.test2.surfconext.nl). Removes another surfconext-ism.


quartje (approved):
Looks nice Thijs, this would also make adding a new app easier. 
I didn't realise this SURFconext'ism was still present, so good riddance to that one :)

PR #370, opened 2021-11-28
Add statefile to /tmp

mvdheenux:
useful for rolling updates

### DIFF
--- a/roles/keepalived/templates/keepalived_loadbalancer.conf.j2
+++ b/roles/keepalived/templates/keepalived_loadbalancer.conf.j2
@@ -38,7 +38,7 @@ vrrp_instance ipv6 {
         {% endif %}
    }
    
-       notify_master "/bin/echo 'master' > /tmp/keepalived.state"
-       notify_backup "/bin/echo 'backup' > /tmp/keepalived.state"
-       notify_fault "/bin/echo 'fault' > /tmp/keepalived.state"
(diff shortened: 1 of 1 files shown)


quartje:
Thanks for the PR!  Since there are two different VRRP instances, wouldn't it make more sense to have different statefiles for each instance? 
And out of curiosity: In what way is this statefile used during rolling updates?

mvdheenux:
Quick response ;). 
Regarding your first question: In most cases, yes. But for this specific use-case I think this isn't adding much functionality. I'll share this playbook with you for reference. 
These statefiles are being used by another Ansible playbook that provides rolling updates by altering Keepalived config. It checks for both statefiles (per host) to contain "backup" before proceeding with installing patches & rebooting. It does this for one host at a time, because time isn't a big concern when patching during the day.

quartje:
I think that you don't get reliable results using the same state file for both instances, since one is always backup and the other always master. Whatever instance writes as last its state to this file decides the state of both. Suppose that the backup instance has written its state, and you want to perform a failover of the master instance  on that host, you wait for the state to be `backup`. But since the statefile already contains `backup` it will not wait.

mvdheenux:
The statefile is located on the host I want to patch, and within that host it's separated between IPV4 and IPV6. So since the statefiles are unique with each host, there should be reliable results. That's what I got back from my own testing at least ;).

quartje (approved):
Seems good to go!

PR #371, opened 2021-11-30
Set umask to 0077 in Mongo and Mariadb backup scripts

quartje:
This changes the umask to 0377, so the produced dumpfiles are not
readable for other users on the system

### DIFF
--- a/roles/galera/templates/mariadb_backup.sh.j2
+++ b/roles/galera/templates/mariadb_backup.sh.j2
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+umask 0077
+
 declare -x PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin:/root/bin
 
 MYSQL_USER="{{ mariadb_backup_user }}"
@@ -60,3 +62,5 @@ rm tmp_sqlend.sql
 
 echo "-- FINISH --"
 fi
+
+umask 0022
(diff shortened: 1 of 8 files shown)


roles/galera/templates/mariadb_backup.sh.j2
```diff
@@ -59,4 +61,5 @@ rm tmp_sqlhead.sql
 rm tmp_sqlend.sql
 
 echo "-- FINISH --"
+umask 0022
```
thijskh, line 64: This is inside the if, while the original setting of the mask is not.

thijskh (commented):
Why the 0377 and not what I would think is common 0077?

Is it not easier/better to just set 0700 on `$FOLDER`/`$backupdir`?

quartje:
> Why the 0377 and not what I would think is common 0077?
> 
To be as strict as possible? 0077 is fine too of course, but why set the execute bit on an archive?

> Is it not easier/better to just set 0700 on `$FOLDER`/`$backupdir`?
Sure, but then the resulting file would still have the original 0022 mask set. If you move that file out of that dir it would still be readable to other users.

thijskh:
But does this even work now? You cannot write to a readonly file?

quartje:
I've tested it and it worked. However, adding an extra +w doesn't hurt at all.
I've changed it, fixed a Galera bug in the process as well. Backups have run already using these scripts. Should be good to ship!

thijskh:
Is the build failing related? It does seem to fail on MySQL/Mongo

quartje:
> Probably needs rebasing on master to make build pass

I thought so too, but alas, the tests themselves are b0rked too. Will fix them

PR #373, opened 2021-11-30
Haproxy: Add the ability to add includeSubDomains; preload to HSTS header

quartje:


thijskh:
I think this is a lot of code and also runtime complexity/evaluation on each request for something that in a normal install will change less than once a year.

What about this approach:

defaults/main.yml:
```yaml
  haproxy_hsts_value: "max-age=34214400"
```

templates/haproxy_frontend.cfg.j2:
```jinja2
http-response set-header Strict-Transport-Security "{{ haproxy_hsts_value }}"
...
http-response set-header Strict-Transport-Security "{{ haproxy_hsts_value }}"
```

As an added bonus an install can then also decide to includeSubdomains but not preload, and you move the max-age value setting to one place from two different places.

quartje:
I've chosen the complicated approach because the loadbalancers serves more than one domain. Your approach would add the header to all the domains you're serving, including all the redirects living on the loadbalancers.

thijskh:
And why would that be a bad thing? The redirects we host are also domains closely tied to the platform.

quartje:
I've reworked it a bit. You might be right in the sense that there is no need for finegrained control over the headers, and having the same header for all outgoing requests makes it more simple

PR #376, opened 2022-02-01
Centralise admin_email instead of using a few different settings.

thijskh:
Also removes another SURF-ism.


quartje (approved):
Nice, I didn;t realise there were so many SURFconext'isms still present. 
I've found two more, and changed them in a commit that I've added to this PR

PR #377, opened 2022-02-15
Update NFS mount options for synced_folders

MKodde:
By forcing NFS version 4 I resolved an issue where my NFS server/client
setup started raising exclusive locking errors.


quartje:
Thanks!

PR #380, opened 2022-03-07
Update lifecycle role to conform to Symfony 5.4

MKodde:
The role is currently set to handle a SF 3.4 project. The recent upgrade to version 4.4 (and the pending SF 5.4 upgrade) require some changes to the lifecycle role.

Most important is the front-controller change. `app.php` no longer exists and is replaced with an `index.php` file. Also, the web folder is no longer the document root. This changed to be the `public` folder.

While at it, I also updated the release script. Removing old, non existing constructions. And replacing them with more correct substitutions.

Disclaimer: I did not test this change. Dear reviewer, would you be so kind to give these changes a test-spin?


quartje:
As discussed earlier, I duplicated this work in another branch, which has been merged. Closing this one.

PR #381, opened 2022-03-09
Direct shib outloggers to engine logout

thijskh:
Which is slightly more informative & platform styled than the
localLogout HTML we shi(b/p).


thijskh:
Compare by clicking logout in Teams on prod (before) and on test2 (after).

quartje (approved):

PR #382, opened 2022-03-22
Lifecycle: 

quartje:
The upgrade to Symfony 4.4 needs some changes for the deployment, as there are other paths for files and configs. I've also included a .env.local in the root of the project to set environment variables

### DIFF
--- a/roles/lifecycle/tasks/main.yml
+++ b/roles/lifecycle/tasks/main.yml
@@ -49,10 +49,16 @@
 - name: Place parameters.yml
   template:
     src: "{{ item }}.j2"
-    dest: "{{ lifecycle_current_release_symlink }}/app/config/{{ item }}"
+    dest: "{{ lifecycle_current_release_symlink }}/config/legacy/{{ item }}"
     mode: 0644
   with_items:
     - parameters.yml
+
+- name: Place .envl file 
+  copy:
+    src: env
+    dest: "{{ lifecycle_current_release_symlink }}/.env"
+    mode: 0644
 
 - name: Create the symfony cache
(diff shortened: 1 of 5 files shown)


roles/lifecycle/tasks/main.yml
```diff
@@ -49,11 +49,17 @@
 - name: Place parameters.yml
   template:
     src: "{{ item }}.j2"
-    dest: "{{ lifecycle_current_release_symlink }}/app/config/{{ item }}"
+    dest: "{{ lifecycle_current_release_symlink }}/config/legacy/{{ item }}"
     mode: 0644
   with_items:
     - parameters.yml
 
+- name: Place .env.local file for local environment overrides\
```
MKodde, line 57: The release script does not leave the .env file in place. It is also in the .gitignore, so it is not cloned into the project when the tarbal is created.

The readme states that the .env.dist file can be used as a template for creating an .env file.

Please note that:
- the env distfile sets an app secret. You might want to ovewrite that one with something randomly generated?
- App debug is set to true by default, not something you want on prod.

Maybe we need to have a small pow wow on what we'd like to track in the repo, and what values should be overridden here?

roles/lifecycle/files/env.local
```diff
@@ -0,0 +1 @@
+APP_ENV=prod
```
MKodde, line 1: See my feedback in the block below. Maybe you need to add some to this template.

MKodde (commented):
Great work so far!

roles/lifecycle/tasks/main.yml
quartje: The secret is set is in the parameters.yml
As we discussed, only a .env file will be put in place now by Ansible

PR #384, opened 2022-04-12
Haproxy: block excessive request on the basis of ip / path / host

quartje:
An ACL was already in place to measure the # of requests to a unique path and host header coming from a unique ip address (the basee32+src fetch method of Haproxy). This change will block an ip address when it hits more than a configured amount of requests per minute (250 in de default setup).

In the future it would be even nice to be able to configure the max rate in a map file (so you don't need to change the whole config file if you want to change the ratelimit)

### DIFF
--- a/roles/haproxy/defaults/main.yml
+++ b/roles/haproxy/defaults/main.yml
@@ -4,7 +4,10 @@ haproxy_metricbeat: False
 # The default http methods whitelist is meant to disable methods like TRACE or DEBUG.
 # OPTIONS is mainly used for CORS requests in OIDC-NG (like SPA apps)
 haproxy_supported_http_methods: "GET HEAD OPTIONS POST PUT DELETE"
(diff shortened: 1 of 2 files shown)


roles/haproxy/defaults/main.yml
```diff
@@ -4,7 +4,10 @@ haproxy_metricbeat: False
 # The default http methods whitelist is meant to disable methods like TRACE or DEBUG.
 # OPTIONS is mainly used for CORS requests in OIDC-NG (like SPA apps)
 haproxy_supported_http_methods: "GET HEAD OPTIONS POST PUT DELETE"
+# maximum number of requests per 10s
 haproxy_max_request_rate: 1000
+# maximum number of requests per host header src ip and path per 1 m
```
thijskh, line 9:
```suggestion
# maximum number of requests per host header, src ip and path per 1 m
```

PR #398, opened 2022-08-24
Haproxy: Allow certain paths to have higher rate limits

quartje:
Rate limits are based on source ip address. Some paths in OpenCoenxt are meant to be accessed with high rates from the same originating IP address. . Like some of the OIDC endpoints. You can now set a higher rate limit for these paths. For now
the ratelimits and paths are static, but if needed in the future they can be made into variables that can be put in the environment variables.


thijskh:
You use map_beg so the entries are matched from the beginning, while the paths mentioned in the map are the path suffixes? So `/oidc/token` will not match `/token`

thijskh (approved):

quartje:
> You use map_beg so the entries are matched from the beginning, while the paths mentioned in the map are the path suffixes? So `/oidc/token` will not match `/token`

Indeed, I should've created a list in the environment vars in the first place ...

PR #399, opened 2022-08-26
Add plugins group to first line

nicodier:


thijskh:
Shouldn't the heading be `[mysqld]`?

See 9f0188925f17291cbab325984834b82f586654be

nicodier:
Actually yes, it should be [mysqld]. I just added a new commit to change that.

PR #404, opened 2022-10-18
TIQR: Add cleanup probability config option

quartje:
This allows you to set a probability that the database session cleanup is triggered.
The default is set to 0.01 (so 1% probability)


thijskh:
Probably better to do this periodically with a cronjob instead of via a probability on some random user’s login?

quartje:
Yes, that is certainly better, but this is already in the code for now, and it works

quartje:
closing this one, as it has been already commited

PR #407, opened 2022-10-28
Replace deprecated include with (include/import)_tasks

thijskh:
`include:` is deprecated and will be removed in Ansible 2.19. Replaced with include_tasks (dynamically, e.g. based on output of earlier task) and import_taks (just statically import file at compile time).

Minimum Ansible requirement is then 2.4.


quartje:
Nice to see this deprecation out of the way. You might want to rebase on master to run the (now fixed) molecule tests

quartje (approved):

PR #409, opened 2022-11-18
Modern spring session has different approach to session cleanup job

thijskh:


roles/teams-server/templates/application.yml.j2
```diff
@@ -76,13 +76,11 @@ voot:
...
 spring:
   session:
-    cleanup:
-      cron:
-        expression: {{ teams_cron_session_cleanup_expression }}
     store-type: jdbc
     jdbc:
       schema: classpath:org/springframework/session/jdbc/schema-mysql.sql
       initialize-schema: always
+      cleanup-cron: "{% if teams_cronjobmaster %}0 13 * * * *{% else %}-{% endif %}"
```
quartje, line 83: Did you change the schedule from once every hour to once a day on purpose (I guess daily is more than enough by the way, just checking)

PR #413, opened 2022-12-09
GHA: Install newer systemd to accomodate for cgroups v2

quartje:
systemd will otherwise fail inside the OpenConect container


quartje:
closing, doesn't work

PR #415, opened 2022-12-16
Feature/upgrade java 11

thijskh:


quartje:
Looks good to me. I did some random deployment tests to check if all goes well. Seems to work fine! Let's merge

PR #416, opened 2022-12-23
Remove java8

thijskh:


quartje (approved):
I've tested it on one of our testing service, seems to be working fine!

PR #417, opened 2023-01-25
Feature/tiqr add apple api http2 proxy

quartje:
The  Apple Push Notification service (APNs) only supports http/2. Since curl shipped with CentOS7 only supports http/1 we are not able to send push messages to this Apple API.
We use haproxy to create a simple reverse http proxy between the Apple API and the TIQR application to  proxy the requests (since Haproxy understands http/2). 
On hosts where the whole stack is installed on 1 machine (and an Haproxy loadbalancer is already present) we add an extra back- and frontend for this purpose.  TIQR can simple push the messages to http://localhost:5000


baszoetekouw:
@pmeulen @quartje Is this still necessary not that we've (almost) moved to containers?

PR #423, opened 2023-03-06
Feature/upgrade mongo 60

quartje:
This is a complete refactor of the Mongo role. It brings the following improvements:
- TLS is always enabled, and automatically deployed, including a CA which is saved in your environment. 
-  Installation of CA signed certificates is done automatically on the hosts
-  Less configuration is needed: only mongo_replication_role needs to be manually set in the host_vars
- Upgrade to Mongo 6.0


tvdijen:
How will you deal with migration, since you can't upgrade MongoDB more than 1 major version at once?

PR #426, opened 2023-04-12
Fix missing TRUSTED_PROXIES environment var in stepup apps.

thijskh:
Name of "engine"_trusted_proxy_ips might be a bit of a misnomer now but it's the correct variable to use.

### DIFF
--- a/roles/stepupazuremfa/templates/env.j2
+++ b/roles/stepupazuremfa/templates/env.j2
@@ -1,3 +1,4 @@
 KERNEL_CLASS='App\Kernel'
 APP_ENV=prod
 APP_SECRET={{ azuremfa_secret }}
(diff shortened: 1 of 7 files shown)


quartje:
If we have an apache that supports it, we might be able to switch to the PROXY protocol in the future, so this config is no longer needed.

quartje (approved):

tvdijen:
> If we have an apache that supports it [...]

I think you can replace it with Apache's remoteip-module, which is generally available.

thijskh:
That is possible but would require to change all apps that do something with this themselves now back to not doing that.

tvdijen:
Maybe I'm thinking to light about this, but all the apps do is catch the X_FORWARDER_FOR-header if it comes from a trusted proxy and replace the IP-address, right?

So if you move that logic to the webserver, your apps will remain working just fine.. All you have to remember is to clean up some code..?

PR #428, opened 2023-05-09
Fix: Gateway SFO uses a different identifier than full 2FA

pmeulen:
E.g:

stepup_uri_self_asserted: 'http://stepup.example.com/assurance/loa1.5'
stepup_uri_sfo_self_asserted: 'http://stepup.example.com/assurance/sfo-level1.5'


thijskh (approved):

MKodde (approved):
Makes perfect sense

pmeulen:
Change is present in master:
https://github.com/OpenConext/OpenConext-deploy/blob/b81b31705096bd3dddfd16f8a597084dd98907d2/roles/stepupgateway/templates/parameters.yml.j2#L103

PR #429, opened 2023-05-17
Make flyway turnoffable

thijskh:


quartje (approved):
That should do the trick
+1 for "turnoffable"

PR #440, opened 2023-08-24
Manage: Make the tabs configurable

quartje:
Manage has tabs that might not be applicable in every situation (e.g oidc, or resource servers, or provisioning). You can disable certain tabs by adding manage_tabs_enabled to your group_vars and overriding the default list


roles/manage-server/defaults/main.yml
```diff
@@ -19,3 +19,10 @@ manage_manage_provision_samlsp_sp_cert: ""
 manage_manage_provision_samlsp_trusted_proxy: false
 manage_manage_provision_samlsp_sign: false
 manage_provision_push: false
+manage_tabs_enabled:
```
thijskh, line 22: In Manage heten dit niet tabs maar `types` geloof ik

thijskh (approved):
Manage will not display tabs/types for which no config file is present.

PR #442, opened 2023-10-25
Increase PHP post_max_size from 8M to 16M.

thijskh:
Manage metadata push to EB can be over 8 MB when you have thousands of entities, so increase it a bit. Cannot be set at runtime for only the push endpoint unfortunately. Do not increase it too much as this would be a DoS risk.


thijskh:
FPM config is indeed better.

quartje (approved):

PR #443, opened 2023-10-26
Haproxy: Cookie based staging environment

quartje:
If you want to add a staging server (to test new features) you can set the cookie staging=true. If that is present,  and you've added a list of servers to the key stagingservers in haproxy_applications that staging server will be used as a backend

We are planning to add the setting of the cookie in Profile and eduID. 
To test, you can set in any Apache configuration of you app:
```
Header always set Set-Cookie "staging=true; path=/; domain=.test2.surfconext.nl; secure; HttpOnly"
```

### DIFF
--- a/roles/haproxy/templates/haproxy_backend.cfg.j2
+++ b/roles/haproxy/templates/haproxy_backend.cfg.j2
@@ -40,3 +40,28 @@
 
 {% endfor %}
 
+{% for application in haproxy_applications %}
+{% if application.stagingservers is defined %}
+#---------------------------------------------------------------------
+# {{ application.name }} staging backend
+#---------------------------------------------------------------------
+#
+  backend {{ application.name }}_staging_be
+  option httpchk {{ application.ha_method }} {{ application.ha_url }} "HTTP/1.0\r\nHost: {{ application.vhost_name }}"
+
+  {%if application.x_forwarded_port is defined %}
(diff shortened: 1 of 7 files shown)


roles/haproxy/templates/haproxy_backend.cfg.j2
```diff
@@ -40,3 +40,28 @@
...
+
+  {%if application.x_forwarded_port is defined %}
+  http-request set-header X-Forwarded-Port {{ application.x_forwarded_port }}
+  {% endif %}
+  http-response del-header ^Strict-Transport-Security:.* #Remove hsts header from backend applications
+  mode http
+  balance roundrobin
+  option httpclose
+
+  cookie HTTPSERVERIDSTAGING insert nocache indirect httponly secure maxidle {{ haproxy_cookie_max_idle }}
```
thijskh, line 60: This is needed so you can have red/blue even for staging servers?

thijskh (approved):

roles/haproxy/templates/haproxy_backend.cfg.j2
quartje: I'm keeping it as close as possible to what we're doing on the production backends.  For me personally it's also handy when debugging, or seeing SAML traces that you're dealing with the staging environment

Technically speaking you're right: We could leave the cookie out since we only have 1 backend.

PR #445, opened 2023-12-11
Reduce privileges for backup- and SST-user

tvdijen:
https://mariadb.com/kb/en/mariabackup-sst-method/#authentication-and-privileges

Only a sub-set is required for SST to function

### DIFF
--- a/roles/galera/tasks/cluster_nodes.yml
+++ b/roles/galera/tasks/cluster_nodes.yml
@@ -215,7 +215,7 @@
   mysql_user:
     name: "{{ mariadb_cluster_user }}"
     password: "{{ mariadb_cluster_password }}"
-    priv: "*.*:ALL"
+    priv: "*.*:GRANT RELOAD,PROCESS,LOCK TABLES,REPLICATION CLIENT ON"
     state: present
     login_user: root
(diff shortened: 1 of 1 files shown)


tvdijen:
I just read that `REPLICATION CLIENT` has become a deprecated alias for `BINLOG MONITOR` in MariaDB 10.6, so depending on the version in use you may want to use the latter.

tvdijen:
According to the [documentation](https://mariadb.com/kb/en/mariabackup-overview/#authentication-and-privileges) the backup-user should have the same privileges.
I've verified this and backup remains in working condition. Could it be that the current privileges for the backup-user are there from the times we used Percona Xtrabackup instead of mariabackup?

tvdijen:
May want to await https://jira.mariadb.org/browse/MDEV-33006 before merging, although everything seems to run smooth. I now have both the repl-user and the backup-user running with just the `RELOAD, PROCESS, LOCK TABLES, BINLOG MONITOR` privileges. Only thing I noticed is that mariabackup is complaining about a missing `CONNECTION ADMIN` privilege, which is why I opened the issue in their tracker.

quartje (approved):
Thanks Tim! I see that the documentation of MariaDB has been updated because of your ticket.

PR #450, opened 2024-04-23
DIYIDP: Use docker for the diyidp. Uses the docker images from Cirrus

quartje:
(https://github.com/cirrusidentity/docker-simplesamlphp)


roles/diyidp/templates/000-default.conf.j2
```diff
@@ -0,0 +1,9 @@
+<VirtualHost *:80>
+    DocumentRoot /var/simplesamlphp/public
+
+    RewriteEngine On
```
thijskh, line 4: Nothing is rewritten

roles/diyidp/templates/saml20-idp-hosted.php.j2
```diff
@@ -49,7 +49,7 @@ $metadata['__DYNAMIC:1__'] = array(
 		// Convert LDAP names to oids.
 		100 => array('class' => 'core:AttributeMap', 'name2oid'),
 		200 => array('class' => 'core:AttributeMap', 'name2urn'),
-		300 => array('class' => 'saml:PersistentNameID', 'attribute' => 'urn:oid:1.3.6.1.4.1.5923.1.1.1.6' ),
+		300 => array('class' => 'saml:PersistentNameID', 'attribute' => 'urn:oid:1.3.6.1.4.1.5923.1.1.1.6', 'identifyingAttribute' => 'urn:oid:1.3.6.1.4.1.5923.1.1.1.6' ),
```
thijskh, line 52: This should probably be instead of `attribute`, not in addition to

roles/diyidp/templates/config-override.php.j2
```diff
@@ -0,0 +1,6 @@
+<?php
+$config['certdir'] = 'config/cert/';
+$config['metadatadir'] = 'config/metadata/';
+$config['auth.adminpassword'] = 'secret';
```
thijskh, line 4: Is this intentional?

roles/diyidp/tasks/main.yml
```diff
@@ -1,127 +1,105 @@
...
-  copy:
+  ansible.builtin.copy:
     content: "{{ diyidp_private_key }}"
-    dest: "{{ openconext_releases_dir }}/simplesamlphp-{{ simplesamlphp_version }}/cert/server.key"
-    owner: "{{ diyidp_fpm_user }}"
-    mode: 0400
+    dest: "/opt/openconext/diyidp/cert/server.key"
+    owner: root
+    group: root
+    mode: "0444"
```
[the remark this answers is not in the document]
quartje: I like to minimalise the ownership, even with the silly read only for root. 
I a later iteration I'd like to use our own containers which are build for overriding the user runs apache, and mapping it to the underlying fs. Sadly, docker does not allow setting different filemodes and ownership of a mounted file in a container

roles/diyidp/templates/config-override.php.j2
quartje: not at all, I missed this. Fixed by using a variable

PR #455, opened 2024-09-19
Only restart docker containers if not new

phavekes:
Restarting a new (freshly started) container might cause issues, e.g. with running migrations. Only restart the container if it's not new.


quartje:
Nice and elegant. Merge & roll

PR #458, opened 2024-11-20
Feature/eb to docker

quartje:
This PR transforms the VM based installation of Docker to a Docker based installation


roles/engineblock/tasks/main.yml.old
```diff
@@ -0,0 +1,288 @@
+---
```
thijskh, line 1: This file accidentally added

roles/engineblock/templates/parameters.yml.j2
```diff
@@ -38,18 +38,17 @@ parameters:
...
     ## or requests (when it acts as an SP) it uses these X.509 certs.
     encryption_keys:
-{% for key, value in engine_keys.items() %}
-        {{ key }}:
-            publicFile: {{ value.publicFile }}
-            privateFile: {{ value.privateFile }}
+{% for engine_key in engine_key_and_certs %}
+      {{ engine_key.name }}:
+        publicFile: "{{ engine_certs_dir }}/{{ engine_key.name}}.crt"
+        privateFile: {{ engine_certs_dir}}/{{ engine_key.name }}.key
```
thijskh, line 50: Nitpick, one is with quotes and the other without for unobvious reasons

thijskh (approved):
The role now no longer executes migrations, so there needs to be some kind of other solution for migrating if this is merged. Otherwise looks good 👍

quartje:
> The role now no longer executes migrations, so there needs to be some kind of other solution for migrating if this is merged. Otherwise looks good 👍

As discussed, a generic role will be made to do migrations, reusable for all symfony apps

PR #460, opened 2024-11-26
Feature/mongo docker

quartje:
This adds a role mongodbdocker so you can install MongoDB on a docker host. 
A seperate docker network is created for the mongo traffic. The apps that use mongo have been adapted so that they can access mongo on a docker host, and that the containers are added to the mongo db docker network using a feature toggle.


baszoetekouw (approved):

baszoetekouw:
@quartje could you please fix he conflicts?

PR #468, opened 2025-01-10
big cleanup. We no longer have PHP and JAVA roles. Everything is in

quartje:
Everything is in docker now.
Removing all the roles related: httpd, php, springboot, java shib and account_gui (which was a leftover).


baszoetekouw (approved):

tvdijen:
Were these files moved somewhere else? Some of this has always been a great reference for me, since SURF always referred to the playbooks as part of their documentation..

PR #470, opened 2025-03-13
Switch to Harica for ACME-based certificates

baszoetekouw:
A number of related changes are in this PR:
 - Move from Sectigo to Harica for certficates (including different acme.sh semantics and options)
 - Switch to issuing a certificate for each domain, instead of one big certificate with lots of SANs  
 - Switch from http-based domain validation to acme-dns based DV.  This works like this:
   - for domain validation, the CA looks for a TXT record in `_acme-challenge.<domain>`
   - the `_acme-challenge` is a CNAME to a fixed entry at an acme-dns service (in our case, acme-dns.surfdomeinen.nl`)
   - `acme.sh` updates the record at the acme-dns-server
   - the CA verifies the TXT record at the acme-dns server.


roles/bind/tasks/create_keys.yml
```diff
@@ -1,15 +1,18 @@
 ---
 - name: create KSK
-  command: dnssec-keygen -K /var/named/keys -f KSK -3 -a RSASHA256 -b 2048 -r /dev/random -n ZONE ha.{{ base_domain }}
+  ansible.builtin.command: |
+    dnssec-keygen -K {{ dns_work_dir }}/keys -f KSK -3 -a RSASHA256 -b 2048  -n ZONE "hb.{{ base_domain }}"
+  creates: "{{ dns_work_dir }}/keys/Khb.{{ base_domain }}.+*"
```
quartje, line 5: Did you test this? It hasn't been used for at least 8 years (the previous change was manually)

roles/bind/tasks/create_keys.yml
```diff
@@ -1,15 +1,18 @@
...
-    stat: directory
-    owner: named
-    group: named
-    mode: 0700
-    recurse: yes
+  ansible.builtin.file:
+    dest: "{{ dns_work_dir }}/keys"
+    owner: "named"
+    group: "named"
+    mode: "u=rwX,go="
```
quartje, line 17: Everywhere else octal mode representation is used. I think it's best to keep it consistent (and personally I think octal representation is more readable, but that's e mere opinion"

roles/haproxy/handlers/main.yml
```diff
@@ -1,43 +1,42 @@
...
-- name: restart rsyslog
-  service:
+- name: Restart rsyslog
+  ansible.builtin.service:
     name: rsyslog
     state: restarted
 
-- name: restart haproxy instances with changed configs
-  service:
+- name: Restart haproxy instances with changed configs
```
quartje, line 17: This one is not used at all anymore.

roles/bind/tasks/create_keys.yml
quartje, line 17: Everywhere else octal mode representation is used. I think it's best to keep it consistent (and personally I think octal representation is more readable)

roles/haproxy/handlers/main.yml
```diff
@@ -1,43 +1,42 @@
...
+  ansible.builtin.service:
     name: haproxy_{{ item.item.name }}
     state: restarted
   when:
     - item.changed
   with_items:
     - "{{ configs_changed.results }}"
 
-- name: restart haproxy redirects instance
```
quartje, line 26: Also obsolete

roles/haproxy/tasks/get_acme_certs.yml
```diff
@@ -1,34 +1,98 @@
...
-- name: Deploy the certificates to haproxy without the need for haproxy to run
-  ansible.builtin.shell:
-    cmd: "DEPLOY_HAPROXY_PEM_PATH=/etc/haproxy/certs   /home/acme/.acme.sh/acme.sh --deploy -d {{ base_domain }} --deploy-hook haproxy"
-  become_user: acme
+# Now deploy the certificates to haproxy
+# We do this in two steps:
+# 1. Cold-deploy the certificates to haproxy's certs directory
+# 2. Make sure haproxy is started, and hot-deploy the certificates to haproxy though the socket API
+
+- name: Cold-deploy the certificates to haproxy  # noqa: no-handler
```
quartje, line 53: What do you mean by "noqa: no-handler"?

roles/haproxy/tasks/get_acme_certs.yml
```diff
@@ -1,34 +1,98 @@
...
-    cmd: "DEPLOY_HAPROXY_HOT_UPDATE=yes   DEPLOY_HAPROXY_STATS_SOCKET=/var/lib/haproxy/haproxy.stats   DEPLOY_HAPROXY_PEM_PATH=/etc/haproxy/certs   /home/acme/.acme.sh/acme.sh --deploy -d {{ base_domain }} --deploy-hook haproxy"
-  become_user: acme
+- name: Hot-deploy the certificates to haproxy  # noqa: no-handler
+  ansible.builtin.command:
+    cmd: |
+      /home/acme/.acme.sh/acme.sh
+        --deploy
+        --deploy-hook haproxy
+        --domain "{{ item }}"
+  environment:
```
quartje, line 89: Ah that looks a lot less ugly than it was

quartje (commented):
Nice work! Two remarks apart from some comments:
1) The handlers now start with a uppercase. All other handlers are all in downcase only. 
2) What's the rationale behind individual certificates in stead of 1 with all SANs in it? It does partly solve the issue that we experienced before: It was not always idempotent because when you request a renewal with the same list of domains which are already present in the certificate acme.sh would create an error.

roles/bind/tasks/create_keys.yml
baszoetekouw: Yes, that's how I found out that the -r flag is obsolete

roles/bind/tasks/create_keys.yml
baszoetekouw: yes, I agree,  Will fix this.

baszoetekouw: Oh now I remember: the problem is the recursion.  If we would use octal notation, all key files would become executable.

roles/haproxy/tasks/get_acme_certs.yml
baszoetekouw: This is to satisfy ansible-lint, which doesn't like `when: foo.changes` tasks.  Generally, those should probably be changd to handlers, but in this case that would complicate the role quite a bit.

roles/haproxy/handlers/main.yml
baszoetekouw: removed.

baszoetekouw:
> Nice work! Two remarks apart from some comments:
> 
>     1. The handlers now start with a uppercase. All other handlers are all in downcase only.

This is fixed now.

>     2. What's the rationale behind individual certificates in stead of 1 with all SANs in it? It does partly solve the issue that we experienced before: It was not always idempotent because when you request a renewal with the same list of domains which are already present in the certificate acme.sh would create an error.

Harica only supports 20 SANs per certificate, so it would need to be split anyway.  The easiest way seemed to just reqeust individual certificates.

PR #481, opened 2025-04-24
Dockerize mujina sp

crosmuller:


pmeulen (approved):
You cleaned up the deployment of the mujina-idp container too, I see. Because you divided the PR in commits I could follow what you were doing, otherwise I would have been to big.
I like the added mujina(-idp/-sp) tags.

PR #492, opened 2025-05-13
Add log rotate configuration for stepup-authentication.log

pmeulen:
Let the rsyslog role write a logrotate configuration for rotating the stepup-authentication.log, like it does for engine


quartje (changes requested):

baszoetekouw:
@pmeulen wat is de status hiervan?  Had jij dit niet al lang gefixed?

Issue #493, opened 2025-05-14
Migrate process_lastseen.php script


pmeulen:
Currently this script is in PHP and requires PHP. The rsyslog role is not compatible with the current Debian target (no yum). 

It is installed by the rsyslog role:
- [process_lastseen.php.j2 ](https://github.com/OpenConext/OpenConext-deploy/blob/2142e48b3d2d42bbec3c534dc0afefea4905aaf1/roles/rsyslog/templates/process_lastseen.php.j2)
- [Installs PHP using YUM](https://github.com/OpenConext/OpenConext-deploy/blob/2142e48b3d2d42bbec3c534dc0afefea4905aaf1/roles/rsyslog/tasks/process_auth_logs.yml#L2-L8)
- [Runs the PHP script from a cronjob](https://github.com/OpenConext/OpenConext-deploy/blob/2142e48b3d2d42bbec3c534dc0afefea4905aaf1/roles/rsyslog/tasks/process_auth_logs.yml#L54-L82)
- Cronjob: [cron_update_lastseen.j2](https://github.com/OpenConext/OpenConext-deploy/blob/main/roles/rsyslog/templates/cron_update_lastseen.j2)

It would make more sense to create a docker image that does all the log processing, i.e including: [parse_ebauth_to_mysql.py](https://github.com/OpenConext/OpenConext-deploy/blob/2142e48b3d2d42bbec3c534dc0afefea4905aaf1/roles/rsyslog/templates/parse_ebauth_to_mysql.py.j2) and [clean_loglogins.j2](https://github.com/OpenConext/OpenConext-deploy/blob/2142e48b3d2d42bbec3c534dc0afefea4905aaf1/roles/rsyslog/templates/clean_loglogins.j2) and also include the stepup-authentication log so logins from direct SFO connections to the steup-gateway other that engine are also seen.

pmeulen:
https://github.com/OpenConext/OpenConext-deploy/pull/496 should address this issue

PR #496, opened 2025-05-15
Feature/rsyslog enhancements

quartje:
Various changes to update the central sysloghost.


quartje:
> Fixes #493 ?

Yes it does. The update of the lastseen table is now done by the Python script that parses the logins to mySQL

crosmuller (approved):

PR #499, opened 2025-05-16
Fix/mongo key change

crosmuller:
- prevent unwanted changing of the ca and possibly all mongo keys and certificates
- add some explanation to the task files
- make sure the role continues in check mode

### DIFF
--- a/roles/mongo/tasks/certs.yml
+++ b/roles/mongo/tasks/certs.yml
@@ -1,6 +1,6 @@
 ---
 # In this task file keys and certificates for the
-# mongo servers are cretae dand signed with the ca from ca.yml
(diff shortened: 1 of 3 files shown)


roles/mongo/tasks/certs.yml
```diff
@@ -1,4 +1,6 @@
 ---
+# In this task file keys and certificates for the
+# mongo servers are cretae dand signed with the ca from ca.yml
```
quartje, line 3: nit: some typos: "create and"

quartje (changes requested):
apart from the typo, looks good to me!

quartje (approved):

PR #500, opened 2025-05-27
rsyslog: Only one forwarding config, allowing to forward to two servers

quartje:
- Add the option to deploy to more than 1 syslog host
- Make 1 logforwarding configuration file, for both Debian and Redhat


crosmuller:
I would put our custom configuration in /etc/rsyslog.d/<FILENAME>.conf, it is cleaner and it wil fix the rhel 9 issue https://access.redhat.com/solutions/7070860

PR #512, opened 2025-06-23
Fix/docker iptables error

crosmuller:
- move all iptables task to the iptables role


roles/iptables/tasks/main.yml
```diff
@@ -15,57 +15,101 @@
...
+        src: ip4dockertables.sh.j2
+        dest: /usr/local/sbin/ip4tables.sh
+        mode: "0750"
+        owner: root
+
+    - name: Disable the default iptables service
+      ansible.builtin.systemd:
+        service: iptables
+        enabled: false
+      when: ansible_os_family == 'RedHat'
```
quartje, line 92: You might want to reconsider if you want to support Redhat based systems (as there is no way to test this, as the SURFconext team does not use them for Docker workloads artm

baszoetekouw: We do still have a number of Rocky systems, which will need to be supported.  Will that work like this?

baszoetekouw: Yeah, looks like it should.

roles/iptables/tasks/main.yml
```diff
@@ -15,57 +15,101 @@
...
-  ansible.builtin.yum:
-    name:
-      - iptables-services
-    state: present
-  when: ansible_os_family == 'RedHat'
+    - name: Disable the netfilter-persistent service
+      ansible.builtin.systemd:
+        service: netfilter-persistent
+        enabled: false
+      when: ansible_os_family == 'Debian'
```
baszoetekouw, line 98: Wouldn't it be better to uninstall the netfilter-persistent package altogether?

baszoetekouw (changes requested):

baszoetekouw (approved):

roles/iptables/tasks/main.yml
crosmuller: Noted

roles/iptables/tasks/main.yml
crosmuller: Noted (for later)

PR #513, opened 2025-06-23
Add script to generate and encrypt new secrets

pmeulen:
Script to generate a random password, secret or hex key of the specified length and optionally encrypt it


crosmuller (approved):
Looks good, I just confused it with gen_secretS.sh so maybe rename one of them?

PR #514, opened 2025-06-25
Feature/sab

baszoetekouw:


crosmuller:
It is  useful to have an "extra"  iptables option, but I would prefer a more generic variable name, it will make the code more understandable for potential non surf users.

baszoetekouw:
Agreed.  But how can we make this generic while at the same time make sure it doesn't conflict with anything else?

For example, we might simple name it `ip6tables_incoming_extra`, but then how can we safeguard this against other applications also using this same name, causing conflicts in the future?

crosmuller (approved):

PR #525, opened 2025-07-16
Update deprecated CI containers to ubuntu-24.04

baszoetekouw:
Ok, molecule tests are running ahgain, now let's get them back working.


crosmuller (approved):

baszoetekouw:
Discussed this with @crosmuller: not sure what the added value of the molecule tests is.  Will try to see if this can get fixed easily, otherwise I'll just disable the molecule tests.

PR #542, opened 2025-09-05
Add  federation metadata cache to azuremfa

phavekes:


roles/stepupazuremfa/tasks/main.yml
```diff
@@ -76,4 +86,7 @@
       - source: /opt/openconext/azuremfa
         target: /var/www/html/config/openconext
         type: bind
+      - source: /opt/openconext/azuremfa/federation-metadata
```
pmeulen, line 89: Better?

`- source:  "{{ current_release_appdir }}/federation-metadata"`

pmeulen (changes requested):
I'm working on related changes in branch-release-591, better to merge this there than to main. Deployment of this version of AzureMFA requires commit e332119fa90314dd01c72064356b8ce06f695a44 which is only in this branch.

PR #558, opened 2025-11-04
Improve ebauth log parsing, and parse stepup-authentication logs also

phavekes:


pmeulen (changes requested):

pmeulen (changes requested):
This filename is strange: `parse_stepupauthauth_to_mysql.py.j2` (two times auth)

pmeulen (changes requested):

roles/rsyslog/tasks/process_auth_logs.yml
```diff
@@ -59,6 +69,16 @@
   with_items: "{{ rsyslog_environments }}"
   when: item.db_loglogins_name is defined
 
+- name: Put log_logins logrotate scripts for stepup
+  ansible.builtin.template:
+    src: logrotate_ebauth.j2
```
pmeulen: @phavekes Is this correct?

pmeulen (approved):
Looks good to me. Now let's test it.

PR #594, opened 2025-12-22
correctly set springboot port aa

crosmuller:
Comment is different from real value


baszoetekouw:
Do we really need to make this configurable?
Afaik this needs to be consistent in the container build only: the Dockerfile and java need to agree, both of which are not handled in Ansible. The correct port is then propagated via Traefik-magick.

baszoetekouw (approved):

PR #595, opened 2026-01-08
change default to irl situation

crosmuller:


baszoetekouw (approved):
After discussion on slack: we will first merge this to bring ansible in sync with the current prod, and then fix prod to bring it up to date to the modern name of the permission.

PR #605, opened 2026-02-03
Re-Repair myconext emails from group_vars

phavekes:
This was fixed in  commit 50715c3121de4dac47ad1369c416d57b072a4237, but was reverted by accident


Tyskai (commented):
<3

PR #638, opened 2026-03-25
Clean up expired invitations

oharsta:
See https://github.com/OpenConext/OpenConext-Invite/issues/681


phavekes:
Already changed in https://github.com/OpenConext/OpenConext-deploy/commit/90b777c599c2203aa6facf0854acea25560c135b

PR #672, opened 2026-05-13
Remove shibboleth deploy config update

ricardovdheijden:
Part of https://github.com/OpenConext/OpenConext-myconext/pull/1198


ricardovdheijden:
Back to draft, needs at least one change in env vars

PR #679, opened 2026-05-26
Haproxy: fix duplicate ssl cert config

baszoetekouw:


baszoetekouw:
The haproxy still serves this cert, which has been obsolute voor over a year: 
Screenshot 2026-05-26 at 10 53 48

baszoetekouw:
This is not a good change, because it breaks acme.sh hot cert pushed via de socket for some reason.  Alternative solution (implemented in [release/663](https://github.com/OpenConext/OpenConext-deploy/pull/715)) is to simply clean up all old certificates.

PR #713, opened 2026-07-17
force renewal certs

crosmuller:


baszoetekouw:
As discussed, this is obsolete because of the use of ARI.  The linting fixes are probably worth keeping though.

baszoetekouw (changes requested):
see above

PR #720, opened 2026-07-30
Release/679

crosmuller:
- create a working filebeat role for our current setup


roles/filebeat/tasks/main.yml
```diff
@@ -1,25 +1,141 @@
...
+          ansible.builtin.file:
+            path: "/etc/apt/keyrings"
+            state: "directory"
+            mode: '0755'
+
+        - name: "Download Elasticsearch GPG key"
+          ansible.builtin.get_url:
+            url: "https://artifacts.elastic.co/GPG-KEY-elasticsearch"
+            dest: "/tmp/GPG-KEY-elasticsearch"
+            mode: '0644'
```
pmeulen, line 43: My preference is to pin the gpg key (i.e. include in our repo).
Now you depend on https trust each time you run the playbook.

roles/galera/tasks/cluster_nodes.yml
```diff
@@ -220,7 +220,7 @@
   mysql_user:
     name: "{{ mariadb_cluster_user }}"
     password: "{{ mariadb_cluster_password }}"
-    priv: "{{ mariadb_cluster_user_privs | default('*.*:RELOAD,PROCESS,LOCK TABLES,BINLOG MONITOR,SLAVE MONITOR') }}"
+    priv: "{{ mariadb_cluster_user_privs | default('*.*:RELOAD, PROCESS, LOCK TABLES, BINLOG MONITOR, SUPER') }}"
     state: present
```
pmeulen, line 224: Does this commit belong in this PR?

pmeulen (changes requested):
How to deal with RPM gpg is something we should standardise for the whole playbook. You decide how to deal with that.

Please do check if the MariaDB change is intended, I did not expect that given the scope of this PR.

roles/filebeat/tasks/main.yml
crosmuller: done

pmeulen (approved):

PR #726, opened 2026-08-17
Merge release branch 681 (Manage)

baszoetekouw:
Only change is fix for https://github.com/OpenConext/OpenConext-manage/issues/690


pmeulen (commented):
How "OpenConext" is this, this seem very SURF specific. Also in Dutch.

baszoetekouw:
That's a good point.  The SURFconext-specific changes to the Manage scheme are historically made is this repo.  Might indeed be better to split those into Openconext-generic and SURFconext-specific business stuff.

I'll make a separate issue for that, to be picked up outside of a release.

